PR #2, opened 2019-02-15
changelog standards

cachedout:


cachedout:
So, there have been no comments on this in the past two weeks. Does that mean this is now in the final comment period?

changelog-format.md
```diff
@@ -0,0 +1,62 @@
...
+The current format for changelog entries at Salt is unstructured and can be difficult to read.
+
+Many users of Salt have complained that they weren't aware of a breaking change which may
+be the result of a poorly structured and inconsistent changelog between releases.
+
+# Design
+[design]: #detailed-design
+
+This RFC standardizes Salt on the 1.0.0 spec of project changelogs outlined at
+keepachangelog.org.
```
waynew:
```suggestion
https://keepachangelog.com/en/1.0.0/
```

changelog-format.md
```diff
@@ -0,0 +1,62 @@
...
+## Alternatives
+[alternatives]: #alternatives
+
+Current alternative is what we have now. ;)
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+Perhaps a sample changelog written for a previous version might
+be appropriate to put together and attach to this RFC?
```
waynew: I think that could be worthwhile 🙂

waynew:
I, for one, am a fan of this idea. I like better documentation.

I don't know if any tools/people expect the existing changelog format, but I'd say that if you're using some tools to automatically process our existing changelog I bet you won't be surprised that something happened :trollface: 

While having a rigorous log format could introduce some pedants, I don't think that's really a bad thing - as the keepachangelog spec says
> Changelogs are for _humans_, not machines.

If someone pops out and says, "Hah! We need to write it thusly instead!" I don't think that's really a bad thing. And yeah, it may require a bit more attention as we shepherd PRs through, but ultimately I think having an improved changelog is a good thing so that effort will be worth it.

waynew:
I think we're still figuring out exactly how this SEP workflow is going, but given the lack of apparent controversy over this suggestion, I'm happy to say that it should go into the final comment period 👍

KChandrashekhar:
@dwoz Since you have been taking care of numbering in file name.

cachedout:
The final comment period on this has passed. :)

KChandrashekhar:
Thanks @cachedout ! Team will review soon

changelog-format.md
```diff
@@ -0,0 +1,62 @@
...
+# Design
+[design]: #detailed-design
+
+This RFC standardizes Salt on the 1.0.0 spec of project changelogs outlined at
+https://keepachangelog.com/en/1.0.0/
+
+Specifically, it provides for the following principles:
+
+- Changelogs are for humans, not machines.
+- There should be an entry for every single version.
```
Ch3LL: Are you proposing that we have just one ChangeLog file with all of the changes for all versions and get rid of each individual release notes file?

Or are you proposing that we include this keepachangelog methodology to each release notes file?

Also is there an example somewhere of someone creating this changelog in an automated fashion from a git log, or would this have to be managed manually? 

also i like this idea overall :) makes a lot more sense for people to consume the information :)

cachedout: Good question. Personally, I think there is a stronger argument for keeping all changes in a single file. I think it's beneficial to have a file where I could, for example, grep for changes to a particular piece of functionality so I can see how it has evolved between versions.

I don't think there is a decent automated approach for changelogs. I've certainly never seen one. IMHO, to do this correctly it must be done manually.

Ch3LL: > keeping all changes in a single file.

Currently I like how much information is added to our feature releases atleast. For example for the 2019.2.0 release we had a breaking change here: https://docs.saltstack.com/en/develop/topics/releases/2019.2.0.html#non-backward-compatible-change-to-yaml-renderer

And we were able to add a lot of additional information with examples of how to work around this breaking change. After reading the https://keepachangelog.com/en/1.0.0/ website it does not seem like with this format we would add all this information (correct me if i'm wrong) to the changelog. What if we keep a changelog in conjunction with at least our feature releases release notes?  In our changelog we could reference the release notes as well.

I think for our point releases the changelog is perfect as right now as you pointed out it is essentially a nicely formatted git log, with both the PRs and issues.

> I don't think there is a decent automated approach for changelogs

i figured this approach would need to be manual just wanted to confirm. We could try to catch it by adding to the changelog when a bug is fixed, feature added during PR reviews as we try to do right now, but would need to be more adamant about it especially bug fixes. Then the release manager would need to review at every release and make sure everything is added.

waynew: I don't see anything in the spec that codifies either accepting or rejecting the idea that changes cannot be links (in fact, there are at least links to user profiles, indicating at least links should be a thing)

I think it would be totally reasonable to do something like this:

```
### Changed
- [Non-backward compatible change to YAML render](https://docs.saltstack.com/en/develop/topics/releases/2019.2.0.html#non-backward-compatible-change-to-yaml-renderer)
- Include JID in minion and master logs
- [Windows `runas` no longer requires password](https://docs.saltstack.com/en/path/to/more/info)
```

Ch3LL: Sounds good :) I'm good with this as it is then.

changelog-format.md
```diff
@@ -0,0 +1,62 @@
+- Feature Name: Standarize on the format of changelogs
+- Start Date: 2019-02-15
+- RFC Status: Draft
+- RFC PR: (leave this empty)
```
waynew:
```suggestion
- RFC PR: https://github.com/saltstack/salt-enhancement-proposals/pull/2
```

changelog-format.md
```diff
@@ -0,0 +1,62 @@
+- Feature Name: Standarize on the format of changelogs
+- Start Date: 2019-02-15
+- RFC Status: Draft
```
waynew:
```suggestion
- RFC Status: Final Comment
```

cachedout:
@waynew Thanks for the review! Your suggestions have been incorporated.

waynew:
Just need one more 👍

0001-changelog-format.md
```diff
@@ -0,0 +1,61 @@
...
+- Versions and sections should be linkable.
+- The latest version comes first.
+- The release date of each version is displayed.
+- Mention whether you follow Semantic Versioning.
+
+Types of changes
+----------------
+- Added for new features.
+- Changed for changes in existing functionality.
+- Deprecated for soon-to-be removed features.
```
KChandrashekhar: Since a critical area most users look for is "Are there any breaking changes in this release", would it make sense to incorporate "Back-ward incompatible functionality" in separate section ? or would that be part of any of the sections listed.

changelog-format.md
KChandrashekhar: Based on @cachedout's proposal about maintaining changelogs in single file, would we go back and change the existing release notes or would we apply this principle only for new/upcoming ones ?

0001-changelog-format.md
cachedout: I'm not sure how "backward-incompatible" and "deprecated" are different. Could you clarify?

KChandrashekhar: Backward-incompatible to me means how you use a service/functionality, or interface with a module has changed but the overall service continues to be offered. For eg.,
https://docs.saltstack.com/en/develop/topics/releases/2019.2.0.html#non-backward-compatible-change-to-yaml-renderer 

Deprecation indicates a service/functionality/something is 'discontinued' now or in future releases. Like no support for Py2 after its EOL. 

Both necessiates user actions to migrate, so I can see why both are similar. 

Since you are well versed in industry notations @cachedout,  I will leave this to you :)

cachedout: This makes sense and I agree with you. I can see how it would make sense to distinguish between deprecation notices which indicate that something *will* be removed/changed and backward compatibility indicates whether the behavior *has* been changed.

waynew: What if we do something like this?

---

## Changes

- **Breaking** - Non-backward compatible change to YAML renderer
- **Breaking** - Frobnosticated the frizzle sticks
- Whirlygigs now have 20% more whirl
- Gew gaws have 15% less gaw, and 30% more gew

Ch3LL:
The `0001-changelog-format.md` SEP has been accepted 

The status in the file has been updated here: https://github.com/saltstack/salt-enhancement-proposals/commit/5150e0aecdd30880499eb8ee88e4949925c48f6a

Thanks!

cachedout:
@waynew and @Ch3LL It might make things easier to understand if we group accepted RFCs into their own directory. It's not obvious to a newcomer that unmerged==unaccepted and committed==accepted IMHO.

Ch3LL:
great suggestion. Technically committed doesn't always mean its accepted, since we can merge if its been rejected. 

From the README:

> At the end of the Final Comment stage, the SEP will be either Accepted or Rejected. Either way, the pull request will be merged.

so i think organizing it in directories would clear that up even more. ping @saltstack/team-core any opinions on organizing these RFCs into directories? I see this being even more useful whne the number of SEPs grows.

KChandrashekhar:
> great suggestion. Technically committed doesn't always mean its accepted, since we can merge if its been rejected.
> 
> From the README:
> 
> > At the end of the Final Comment stage, the SEP will be either Accepted or Rejected. Either way, the pull request will be merged.
> 
> so i think organizing it in directories would clear that up even more. ping @saltstack/team-core any opinions on organizing these RFCs into directories? I see this being even more useful when the number of SEPs grows.

I remember seeing organization by folders in other open source projects too. I'm definately up for it.

waynew:
I'm +1 on having rejected SEPs in their own directory, +/-0 on accepted ones (I'm not against it, but I'm not particularly for it).

I've thought about the idea of either having the wiki on here or add an index file that contains an easier to digest list/version of things.

Though now that I'm writing about it, perhaps it does make sense to have accepted/rejected, as those are the two "final" states (there's implemented, up for grabs, obsolete, and replaced, but I don't think there's nearly as much reason to pull those into their own directories. If we needed to do something like that I'd recommend the [TagSpaces](https://www.tagspaces.org/) approach of just doing something like `0001-changelog-format[implemented].md` or something).

cachedout:
@waynew The problem, I think, is here:

> At the end of the Final Comment stage, the SEP will be either Accepted or Rejected. Either way, the pull request will be merged.

Given that, how does one look at the repo either from a checked-out copy or just from the main page of this repo and tell which is which?

waynew:
`grep -Inr accepted .` 😂 

(which is the point behind accepting even rejected PRs - so one doesn't have to have GitHub available to determine that we've found proposal X to not be appropriate for Salt)

Yeah, I think I'm +1 on having an accepted/rejected directory. I think it makes sense that when one creates a PR they just create the file in the root of the repo. The acceptance (or I guess rejection) workflow then should be:

1. Voting happens
2. Votes tallied, Accepted/Rejected label is applied
3. Author (or core team member) adds accepted to the status, moves the file to the "accepted" dir, and pushes that change
4. PR is merged
5. ???
6. Profit!

cachedout:
@waynew I think that makes a lot of sense. Excellent!

PR #6, opened 2019-03-19
SEP 3 - SEP statuses

waynew:
View SEP [here](https://github.com/waynew/salt-enhancement-proposals/blob/SEP-statuses/0003-more-sep-statuses.md), and the [updated README](https://github.com/waynew/salt-enhancement-proposals/blob/SEP-statuses/README.md)


waynew:
Time to enter the final comment phase

PR #8, opened 2019-03-25
SEP 5 - Retiring Python 2 Support

waynew:
View [the SEP](https://github.com/waynew/salt-enhancement-proposals/blob/py2-deprecation/0005-retire-py2-support.md)


py2-deprecation.md
```diff
@@ -0,0 +1,134 @@
...
+page](https://pypi.org/project/pyOpenSSL/#history) to identify the which
+versions of Python are supported by pyOpenSSL.
+
+### Note 3
+
+A future SEP will define a schedule for ongoing depreciation of Python versions as they also reach their end-of-life.
+
+### Note 4
+
+A future SEP will define the salt-ssh Python deprecation schedule.
```
gtmanfred: Does this mean that salt-ssh will be deprecated or that salt-ssh will continue to support python2 (which means all modules and states will still need to support it) until some point in the future?

py2-deprecation.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Actions
+[Actions]: #Actions
+
+SaltStack users will see the Python 2 deprecation implemented in the next few
+releases as follows: 
+
+1. Future releases of SaltStack will have a Python 2 deprecation warning.
+2. SaltStack will create packages that will allow for a smooth transition for users.
+3. In accordance with the schedule decided on in this SEP, testing for Python 2
```
gtmanfred: Will this be turned off for salt-ssh as well? even though if your remove python2 support for modules it will end it being able to be tested?

How will we make sure that salt-ssh still works against python2, since only the python2 tests run salt-ssh against python2 on the remote.

waynew:
> This all seems great, i just would like some more information around salt-ssh.

I'm not sure when the salt-ssh SEP will be released, but that's when we're going to address the concerns about salt-ssh - I'm pretty sure @Ch3LL is working on those.

Ignoring salt-ssh on the part of *this* SEP is intentional - salt-ssh is likely going to have a bunch of edge cases, or thorny issues that are all orthogonal to dropping Python2 support in Salt. IIRC all non-EOL'd supported OSes at the estimated time of the Sodium release will have Python3 support of some sort (with a minimum version of 3.4, I think). What the salt-ssh story looks like at that point will likely involve plenty of discussion.

Hence, the forthcoming SEP on salt-ssh 🙂

gtmanfred:
> Will this be turned off for salt-ssh as well? even though if you're remove python2 support for modules it will end it being able to be tested?
> How will we make sure that salt-ssh still works against python2, since only the python2 tests run salt-ssh against python2 on the remote.

I will be a :-1: on this SEP until it can be determined how turning off these tests will affect salt-ssh, because the python2 tests for salt are the only way python2 + salt-ssh is tested.

And i do not appreciate my questions being marked as resolved, just because you feel like they are being resolved.  Changes in this SEP have real effects on other peoples environments, the person who proposed the question should have the finally say on if it was resolved or not.

waynew:
> I will be a 👎 on this SEP until it can be determined how turning off these tests will affect salt-ssh, because the python2 tests for salt are the only way python2 + salt-ssh is tested.

To clarify - are you:

- against the Python 2 depreciation?
- against no longer testing on Python 2?
- separating salt-ssh's deprecation SEP from the general Python 2 deprecation SEP?

I realize that you have concerns about salt-ssh and testing. So do I.

I **do not** at all intend to dismiss them as invalid concerns, *except as pertains to this SEP*. Exactly how we handle salt-ssh is a question that this SEP deliberately has no answers for because the answers (and questions) are too important to be conflated with the equally important question of "how do we deprecate Python2?"

From my point of view, this SEP is *only* about depreciating Python2 within Salt. From the perspective of everything that is *not* `salt-ssh` (and yes, I realize this is a totally imaginary distinction, since salt-ssh is barely anything on top of everything else Salt 😛but bear with me)

My hope is that you can say, "Ah, yes, I see. Dropping Python 2 support is a good thing for Salt and its users, and I support that! I have questions about how salt-ssh is going to work going forwards because (well, a lot of reasons), and I will make sure to raise those on that upcoming SEP - I'm especially concerned that salt-ssh continues to get tests as long as it supports Python 2."

---

Nothing in this SEP should be construed to have any bearing of any kind on what we do with `salt-ssh`, and it feels like a deliberate misreading of this SEP to feel otherwise. I apologize if I miscommunicated my intent and feelings.

Maybe that's because I wasn't clear enough in this phrase:

> A future SEP will define the salt-ssh Python deprecation schedule.

If that doesn't make it clear that a future SEP is entirely responsible for salt-ssh's depreciation schedule, including how it's tested, deployed, etc., I'm more than happy to update that wording 🙂

gtmanfred:
My concern is that this test lays out disabling python2 tests for salt-ssh, without a way to test salt-ssh.

I am perfectly fine with handling all other salt-ssh deprecations in a different SEP for salt-ssh. except for this one point.  If we disable the python2 tests whole sale without making a change for how the salt-ssh tests run, we will be disabling all the tests for salt-ssh.  This should be a self contained change.  If we want to disable python2 tests, but still support salt-ssh with python2, then we need to test salt-ssh with python2.

By disabling the python2 tests, you are effectively ending all testing of python2 on salt-ssh.

If we just include one more step to allow to specify which python version salt-ssh runs under, and include that in the python3 test suite to run a salt-ssh minion vs python2, that will move me from a -1 to a +1.

py2-deprecation.md
```diff
@@ -0,0 +1,134 @@
...
+   will be turned off.
```
waynew:
```suggestion
   will be turned off. (Note: This makes no assumptions about salt-ssh testing
   which will be specified in a future SEP - see [note 4](#note-4))
```

@gtmanfred does this satisfy?

gtmanfred:
Alternatively, removing all mention of disabling tests from the proposal would also get me to a 👍 
Since that is inherently understood to be a thing that we would cross when we get to it, but is not needed in a broad proposal on how to move away from supporting python2.

py2-deprecation.md
gtmanfred: I would prefer just removing all mention of disabling tests, since if we aren't supporting python2, having tests for it would be redundant.

And we could potentially still have python 2 tests running for salt-ssh.

gtmanfred: It would also make the SEP more broad and more define a direction to go instead of a rigid path to follow.

Ch3LL:
Just as an update, yes I'm currently working on the Salt-SSH RFC. It is taking me a bit longer than expected because I wanted to test a couple of ideas about the proposal around building python3. Without getting into the weeds of the proposal as I would like to keep the comments to the salt-ssh RFC when i publish it, the proposal will include building out a minimal python3 binary. I wanted to build and test out static vs dynamic builds of python3 first. Apologies on the delay, i hope to get some time to test these out this week.

gtmanfred:
@mattp- @cachedout y'all commented on the last versions of this, any opinions here?

mattp-:
Generally looks good to me 👍

py2-deprecation.md
```diff
@@ -0,0 +1,132 @@
...
+
+In light of Python 2.7 reaching its end of life (EOL) on Jan 1st 2020, Python 2
+will be deprecated from SaltStack no earlier then the Sodium release, that is
+either the Sodium release or a later release.
+
+
+# Motivation
+[motivation]: #motivation
+
+Python 2.7 will reach it’s official end of life (EOL) on [January 1st,
```
cachedout: `its`

waynew: And apparently a bunch of other smart quotes. Fixed now ✨

waynew:
I believe that it's time for this SEP to enter the final comment period

waynew:
And with that, the aye's have it. Thank you Python2 for being an amazing language. It's not you, it's me. Me and Python3 :trollface:

PR #9, opened 2019-03-26
SEP 6 - Remote client RFC

cachedout:
This has been migrated from the original repo:

https://github.com/saltstack/salt/pull/50976

*IMPORTANT* Please read the comments in the old issue before commenting
on this one.


cachedout:
Two week bump. :)

waynew:
It's amazing how much coming back to a thing after you've slept a few times helps 😂 

I'm definitely a fan of the *idea* - as to the underpinnings/implementation details, I definitely want/need to grok that a bit more.

My main concern is hypocratic - "first, do no harm" - I want to make sure that whatever we come up with is not going to add (significantly?) to the attack surface for anyone using this feature. 

Other than that, I don't know enough about the internals of Salt to have other worries :trollface: 

> This might well make that problem worse though and the handling of "duplicate" returns would need to be addressed more thoroughly.

To me it seems like it should be trivial to do some kind of dict merging where the returns are by master.

Actually, it may be that as prerequisite to the implementation of this SEP we must update returns to work in a simple/elegant fashion...


Anyway, to sum up, I like the idea, I think fundamentally it's a good one. I think implementation will be tricksy business.

remote_client.md
```diff
@@ -0,0 +1,130 @@
...
+syndics. This might well make that problem worse though and the handling of
+"duplicate" returns would need to be addressed more thoroughly.
+
+If adopted, we would need to determine whether to remove the `root_key`
+system or whether to allow it to continue as an option.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+First and most obviously, an administrator might now wish to allow remote
```
dafyddj:
```suggestion
First and most obviously, an administrator might not wish to allow remote
```
That should be a negative, right?

remote_client.md
```diff
@@ -0,0 +1,130 @@
...
+ * Enhancing support for a master to be controlled via events on the master event
+   bus. This might make integration with a reactive architecture substantially
+   more robust.
+
+ * Simply creating a special kind of minion that has rights on the existing
+   minion-publish system. However, in the author's view, the overhead of this
+   in terms of performance and start-up speed for a client would be too severe.
+
+ * Create a unified client which can publish commands seamlessly between salt-api
+   instances and local salt masters. This could potentially be extetended to
```
dafyddj:
```suggestion
   instances and local salt masters. This could potentially be extended to
```

PR #11, opened 2019-05-01
SEP 8 - Python 3 Only Salt-SSH

Ch3LL:


waynew:
This SEP will be number 8, feel free to update your filename and SEP 🙂

python3-ssh-template.md
```diff
@@ -0,0 +1,83 @@
+- Feature Name: Python 3 Only Salt-SSH
+- Start Date: 2019-05-01
+- SEP Status: Draft
+- SEP PR: (leave this empty)
```
waynew:
```suggestion
- SEP PR: https://github.com/saltstack/salt-enhancement-proposals/pull/11
```

0008-python3-ssh-template.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Design
+[design]: #detailed-design
+
+The design proposal to allow Salt-SSH to work with Python 3, includes multiple options:
+  1. By default, if python versions match we will simply use the system python and Salt-SSH
+     will run as it always has.
+  2. Add a Salt-SSH configuration (for example: pre_flight) to run raw commands before the
+     Salt-SSH command. This will allow a user to create a custom script they can run to install
+     Python 3. It will only run these pre_flight commands if the tarball is not copied over.
```
gtmanfred: :+1:

0008-python3-ssh-template.md
```diff
@@ -0,0 +1,83 @@
...
+     if they always want to use the binary option.
+
+If the python versions do not match when Salt-SSH is run a message will be printed out asking the user
+if they want to copy over the binary, and informing them of the other alternative options pre_flight and
+ssh_ext_alternatives.
+
+
+How are we going to build the Python 3 binary?
+We will build the binary statically for both x86_64 and ARM architecture. We will include the x86_64
+binary by default in the Salt-SSH package. If the ARM architecture is detected we will include a warning
```
gtmanfred: As long as there is a way to upgrade this, like you can with salt-cloud -u for the bootstrap script, i think this is fine.

0008-python3-ssh-template.md
```diff
@@ -0,0 +1,83 @@
...
+[alternatives]: #alternatives
+
+Alternatives for Python 3 static binary:
+- Use an already maintained minimal version of Python 3, but we will not have as much control over what is added into the binary.
+- Use a dynamic build, but this will require building out a package per distro.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+When we build the static python build we will need to review the licenses of all the libraries included in the build.
```
gtmanfred: It would be good to have possibly a dashboard or an easy way for salt to be alerted about any CVEs in dependencies.

This might be really easy to do using github and their security alerts

https://help.github.com/en/articles/about-security-alerts-for-vulnerable-dependencies

or using pyup.

https://pyup.io/safety/

0008-python3-ssh-template.md
Ch3LL: Great suggestion :) Added here:

https://github.com/saltstack/salt-enhancement-proposals/pull/11/commits/7cda959762ccd3630af1abb34621489170ed5d59

0008-python3-ssh-template.md
Ch3LL: I added this to the `Python Binary Security Releases:` paragraph instead as i thought that made sense.

https://github.com/saltstack/salt-enhancement-proposals/pull/11/commits/16dc67f57dba929be335dfcae0649a48529211b8

Akm0d: Have we ever considered something like [PyInstaller](https://pypi.org/project/PyInstaller) to package a binary with python and all its dependencies?

meaksh:
Just as a note here. We're currently solving this situation in a different way based on the changes that were introduced here: https://github.com/saltstack/salt/pull/46684/

At SUSE, we're still dealing even with old Python 2.6 systems and we use `salt-ssh` from a Python3 master to deal with them.

In order to do that, we provide a `py26-compat-salt` package [1] (based on 2016.11.10) which is installed on the master as an alternative salt codebase (outside /usr/lib/) and we provide also the necessary `ssh_ext_alternatives` settings [2] in order to use that codebase when dealing with Python 2.6 in the target system.

What I like from the approach that RFC provides is that this will keep all functionality & features that the latest Salt (based on Python3 and installing on the master) when targeting a system with an older Python version, instead of having to different codebase and do backporting or arrange SLS to keep the compatibility.

On the other hand, I'm a bit worried about the problems with other Python dependencies (apart from SaltSSH) that might be required from the Python3-based execution & state modules that are been executed as part of the particular action that the `salt-ssh` call is performing and are not available on a Python2 targeted system.

IIUC, those libraries required by any execution & state modules that is going to be executed would need to be added as `ssh_ext_alternatives`, right?

[1] - https://build.opensuse.org/package/show/systemsmanagement:saltstack:products/py26-compat-salt
[2] - https://build.opensuse.org/package/view_file/systemsmanagement:saltstack:products/py26-compat-salt/py26-compat-salt.conf?expand=1

Ch3LL:
> IIUC, those libraries required by any execution & state modules that is going to be executed would need to be added as ssh_ext_alternatives, right?

If the user is using `ssh_ext_alternatives` then yes they would need to define the libraries there. My understanding is that this is how it currently works.

If they are using a different approach such as the python3 system library,  they can use a `pip.install` state/module to install the libraries similar to how users would currently do so.

PR #15, opened 2019-06-12
SEP 11 - Windows: copy python.exe to salt-*.exe

markuskramerIgitt:
https://github.com/saltstack/salt/issues/53254


waynew:
This will be SEP 11 - feel free to update your .md file!

twangboy:
The problem with just renaming `python.exe` to `salt-minion.exe` is that many other salt tools use `python.exe`. For example, `salt-call`, `salt-master`, `salt-key`, `salt`, etc... We would probably want to create binaries for each of these.

markuskramerIgitt:
@twangboy This SEP is not about renaming python.exe, but copying python.exe.

damon-atkins:
> For maintenance and bug-reporting, it would be beneficial if the salt client processes in the Windows task manager would be named

image

What problem are you trying to solve?
Do you have any examples?

waynew:
I *think* @twangboy's point is that we'd want to have several copies, such as `salt-call.exe`, `salt-master.exe`, `salt-key.exe`. I need to get on a Windows box, but I seem to recall that using `[entrypoints]` with setuptools will generate `.exe` files for each command. Of course, Salt isn't using setuptools, but it's possible that we might be able to vendor in that sort of behavior.

markuskramerIgitt:
@damon-atkins, the problem is that most of our customers, and probably many support colleagues are unaware of the 'Comand line' column you show. At least I was until now. Customers are neither trained nor willing to dig into the task manager. They see the 'Name' column and report a problem with 'Python.exe', which increasingly gets connected with the ominous "Salt" thing. 
As the connection is not obvious, more not-obvious connections from problems to Salt are created. So I have an increasing number of "my computer is slow" problems, which must be the "Salt" thing. 

Any Programs I know has its name in the 'Name' column. Salt does not. I find this a little sad for such a good program, and I have an increasing stream of unrelated problems, or bad reputation.

Btw, how did you include the picture?

markuskramerIgitt:
@twangboy, @waynew I agree: any salt-* command should have its own *.exe.

I think we can achieve that with a simple Windows `copy`.

I also would like that the Wix tool, after the Nullsoft tool, finally becomes the second installation procedure.
A third procedure (setuptools), would not help in that direction.

waynew:
I can confirm that yes, the setuptools approach does create a `.exe`.

I think what we would want to do then is do whatever setuptools is doing to create the .exe on Windows.

0011-copy_python_exe.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+For maintenance and bug-reporting.
+
+# Design
+[design]: #detailed-design
+
+copy python.exe to salt-minion.exe and reference the latter, e.g in salt-minion.bat
```
waynew: There may some better alternatives than strictly copying python.exe, like using the Setuptools approach. I think that the SEP should just specify that we want to create `salt-*.exe`, and provide a couple of options, since it's likely that there will be issues with one approach or another.

waynew:
@markuskramerIgitt I think fundamentally we all like the idea, though I think we should be flexible on the implementation.

thatch45:
I will throw a wrench in the works, I recently got pyinstaller working reliably with Salt and it could make a single salt.exe that could encapsulate all salt commands and make it so that we only need to distribute a single file. Let me chat with @twangboy about this idea.

waynew:
Sounds like we definitely have some options on how we get it done, but the fundamental idea is a good one 😁

markuskramerIgitt:
I continue to suggest that the Windows task manager shows specific names for the processes:

- `salt-minion`
- `salt-master`
  
I no longer suggest how to achieve this goal,  e.g. by copying python.exe or creating several *.exe like Thomas suggests.

0011-copy_python_exe.md
markuskramerIgitt: I no longer want or can discuss the implementation. Please feel free to change the detailled design.

thatch45:
We did try this out and it looked like the executable was making an additional process which would arguably negate ethe benefit.

thatch45:
yes, it does, unfortunately. I will try to get something inline with salt-bin to test

markuskramerIgitt:
@thatch45 , `c:\salt\bin\python.exe` does not contain the word "python" to call a spawned process "python.exe". 
I would bet another script was calling the original python.exe. 

We get increasing heat inside the company from "python.exe"

0011-copy_python_exe.md
markuskramerIgitt: I will try to find time for a test

thatch45:
I will need to comment that at this time all Linux distros show up in the process table as `python salt-minion` and that the complaint that you are surfacing does not seem to be affecting others.
If you want to call the salt-minion.exc directly you can do so. Or you are more than welcome to try and create a single binary using a tool like salt-bin: https://github.com/saltstack/salt-bin

markuskramerIgitt:
`python salt-minion`, as on Linux, would give Windows end users or the help desk enough information to route the problem ticket.

I will now stop working on this SEP until someone else joins who is affected too and would like to collaborate. 
Please feel free to close this SEP at your discretion.


I could have another look into salt-bin, once we are able to use Python3.

waynew:
tl;dr: `salt` showing up in the process table 👍 - copying `python.exe` to `salt-*.exe` 👎 

---

As a follow-up, we're rejecting this not because we dislike the idea of having `salt` show up in the process table - just the implementation of copying python.exe to salt.exe.

Also, we do have some plans around using salt-bin, which would accomplish the desired goal here.

If you'd like to open a new SEP with a different implementation based on the discussion here, that would be 👍 or we can just let things resolve naturally.

Fundamentally *we like the idea of having Salt show up in the process table*, we just don't like this specific implementation.

PR #16, opened 2019-07-16
SEP 12 - Engines live configuration

cbosdo:
Allow adding /removing engines without restart the minion

### DIFF
--- a/update-engines-without-minion-restart.md
+++ b/update-engines-without-minion-restart.md
@@ -1,0 +1,37 @@
+- Feature Name: update-engines-without-minion-restart
+- Start Date: 2019-07-16 
+- SEP Status: Draft
+- SEP PR: https://github.com/saltstack/salt-enhancement-proposals/pull/16
+- Salt Issue: 
+
+# Summary
+[summary]: #summary
+
+Being able to restart minion engines without having to restart the minion itself.
+
+# Motivation
+[motivation]: #motivation
+
+As a user I need to run a new engine on a minion, but restarting the minion needs to be done carefully and may not happen when I need it.
+This is already possible with beacons.
+
+# Design
+[design]: #detailed-design
+
+The engine configuration could be either in the configuration or in the pillar.
+When refreshing the pillar, loop over the engines configuration and start or stop the engines according to the new configuration.
+
+## Alternatives
+[alternatives]: #alternatives
+
+The alternative is to manually restart the salt minion, with all the problems this can bring.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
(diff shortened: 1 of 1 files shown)


update-engines-without-minion-restart.md
```diff
@@ -0,0 +1,37 @@
+- Feature Name: update-engines-without-minion-restart
+- Start Date: 2019-07-16 
+- SEP Status: Draft
+- SEP PR:
```
waynew:
```suggestion
- SEP PR: https://github.com/saltstack/salt-enhancement-proposals/pull/16
```

cbosdo: Done. Thanks

waynew:
@cro do you have any thoughts about this? Someone suggested you might be familiar with this.

thatch45:
I think this would be great!

max-arnold:
+1 for this feature!

Recently I wrote a couple of beacons that take some time to run, and unfortunately, they block the minion event loop.  An engine would be a better fit for this task; however, it is not so easy to configure and deploy/update.

My first concern is the automatic restart on each pillar refresh. I'm not sure when it happens, and there is a related bug https://github.com/saltstack/salt/issues/54941. This could lead to unnecessary restarts, even if there are no actual changes in engine code or configuration.

Also, an engine could have some internal state that needs to be persisted/flushed, so maybe it makes sense to provide a pre-shutdown/restart callback (or a signal handler). On first glance, forced restarts could be problematic for some of the [existing engines](https://github.com/saltstack/salt/tree/master/salt/engines).

max-arnold:
And I just realized that engines could be run on a master, so it is also useful to restart them without bringing the master down. How could this be handled?

waynew:
> I think this would be great!

@cbosdo not much more to say than that.

We discussed this in our core meeting today, so the rest of the approvals should show up soon and we can get that merged in.

@garethgreenaway has some relevant information about this that he can provide, but he's pretty busy right now so it may be January before he can get to this.

thatch45:
By relevant information, this means that he can outline what it will take to get this added in. Likely a runner for master engines and an execution module for minion engines. Just because we merge a SEP does not mean that we will be able to implement it, a SEP is an approval of a feature so someone can feel safe dedicating the time to create it :)

cbosdo:
@thatch45 @waynew thanks for the infos. I am also rather busy with the virt module and states as well as other related things. Wouldn't be able to start on that before January either.

waynew:
@cbosdo This SEP has been accepted 👍 When you (or anyone else) has time to start working on this, we look forward to seeing your implementation!

cbosdo:
Thanks @waynew. I'm still fairly busy here, if someone wants to pick it feel free to do it. I'll try to come to it later, most probably around spring

PR #21, opened 2019-10-04
SEP 15 - Blacken Salt :black_heart:

waynew:
See the [text of the SEP](https://github.com/waynew/salt-enhancement-proposals/blob/black/black.md) 🖤

### DIFF
--- a/black.md
+++ b/black.md
@@ -37,6 +37,13 @@ the use of Black and lint. As `pre-commit` runs `black`, `isort` and lint
 checks, it will allow us all to focus on the logic and design of the PRs
 which will increase the stability and quality of Salt.
 
+#### Lint Names
+
+Additionally, on the rare occasion that we need to disable lint rules, we will
+disable based on name, instead of number. The existing disabled lint numbers
+will be changed to the corresponding names as part of the PR that implements
+this SEP. 
+
 ### Git Blame? 
 
 Because the introduction of Black will absolutely touch every file in the
--- a/black.md
+++ b/black.md
@@ -33,7 +33,7 @@ there is only “one way to do it”. Accepting double quotes was extremely
 difficult for Tom, but if he can do it, anyone can.
 
 Salt will also adopt [pre-commit](https://pre-commit.com) in order to automate
-the use of Black and lint. As `pre-commit` runs `black`, `isort` and lint
+the use of Black and lint. As `pre-commit` runs `black`, `isort` (using `force_single_line=True`) and lint
 checks, it will allow us all to focus on the logic and design of the PRs
 which will increase the stability and quality of Salt.
 
--- a/black.md
+++ b/black.md
@@ -1,0 +1,74 @@
+- Feature Name: blackened-seasoning
+- Start Date: 2019-10-04
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Adopt `pre-commit`, `black`, and `isort` for Salt.
+
+# Motivation
+[motivation]: #motivation
+
+There have been a *number* of PRs to Salt that have had minor style changes
+requested, either to bring the coding style in to line with the general Salt
+project or within the specific module, when it deviates from Salt's typical
+standards.
+
+Additionally there are a number of commits that are produced just to satisfy
+the linter. If we automate the formatting of our code, this goes away. We also
+achieve consistency throughout the Salt codebase.    
+
+
+# Design
+[design]: #detailed-design
+
+### Black, Pre-commit, and You
+ 
+Salt will adopt [Black](https://github.com/psf/black) as our code formatter of
+choice. By adopting an opinionated formatter we are taking the approach that
+there is only “one way to do it”. Accepting double quotes was extremely
+difficult for Tom, but if he can do it, anyone can.
+
+Salt will also adopt [pre-commit](https://pre-commit.com) in order to automate
+the use of Black and lint. As `pre-commit` runs `black`, `isort` and lint
+checks, it will allow us all to focus on the logic and design of the PRs
+which will increase the stability and quality of Salt.
+
+### Git Blame? 
+
+Because the introduction of Black will absolutely touch every file in the
+codebase (except perhaps a `__init__.py` file or two), a member of the Salt
+core team will create that commit using the `--author` flag set to `Salt
+Refactor <jenkins@saltstack.com>` so it’s
+obvious in `git blame` that those changes are due to `black`ening the codebase.
+
+This commit hash will also be added to a `.git-blame-ignore-revs` file, so it
+can easily be used in the `git blame --ignore-revs-file=.git-blame-ignore-revs`
+command, as well as automatically be used with the `git hyper-blame` extension.
+
+### Updated Docs
+
+Adopting these tools for automation will require us to update our documentation
+to make it simple for Salt contributors.
+
+## Alternatives
+[alternatives]: #alternatives
+
+We could use `sq-black`, a fork of Black that allows single quotes. This
+would require maintenance of the fork.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+None?
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We could *not* do this, because it will add some overhead when researching
+issues via `git blame`. Especially formatting changes. It will require updating
+the documentation. There will also require rebasing of any work done before the
+formatting change.


black.md
```diff
@@ -0,0 +1,74 @@
...
+### Updated Docs
+
+Adopting these tools for automation will require us to update our documentation
+to make it simple for Salt contributors.
+
+## Alternatives
+[alternatives]: #alternatives
+
+We could use `sq-black`, a fork of Black that allows single quotes. This
+would require maintenance of the fork.
```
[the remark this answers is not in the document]
s0undt3ch: Though sticking with the original black project will reduce headaches in the future.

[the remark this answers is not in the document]
cheburakshu: `black -S foo.py` doesn't format strings and let you use single quote. But, the codebase will have single and double quotes.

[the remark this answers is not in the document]
s0undt3ch: Oh no. It's all or nothing, no middle ground. Either single, or double.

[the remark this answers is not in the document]
cheburakshu: ok. 👍

thatch45:
This has been a long hard fight to use single quotes, but this seems to be the end. I see how this will solve many other formatting issues which is a worthy trade-off.

black.md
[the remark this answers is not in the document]
mirceaulinic: FWIW, Black has an option `--skip-string-normalization` that doesn't enforce double or single quotes. Might be a good trade-off, if we want to preserve the current style. :man_shrugging:

thatch45:
`--skip-string-normalization`

black.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Black, Pre-commit, and You
+ 
+Salt will adopt [Black](https://github.com/psf/black) as our code formatter of
+choice. By adopting an opinionated formatter we are taking the approach that
+there is only “one way to do it”. Accepting double quotes was extremely
+difficult for Tom, but if he can do it, anyone can.
+
+Salt will also adopt [pre-commit](https://pre-commit.com) in order to automate
+the use of Black and lint. As `pre-commit` runs `black`, `isort` and lint
```
nixjdm: I recommend [https://github.com/asottile/reorder_python_imports](reorder_python_imports) instead of isort. It shares black's philosophy of being nearly unconfigurable.

If you want to use isort, I would also suggest using reorder-python-imports's choice of breaking imports into individual lines by setting `force_single_line` to `True`. This will achieve [reorder-python-imports's stated single aim](https://github.com/asottile/reorder_python_imports#why-this-style) of having a practical effect beyond just style - avoiding merge conflicts. With the number of Salt PRs, that could really help.

black.md
```diff
@@ -0,0 +1,74 @@
...
+checks, it will allow us all to focus on the logic and design of the PRs
+which will increase the stability and quality of Salt.
+
+### Git Blame? 
+
+Because the introduction of Black will absolutely touch every file in the
+codebase (except perhaps a `__init__.py` file or two), a member of the Salt
+core team will create that commit using the `--author` flag set to `Salt
+Refactor <jenkins@saltstack.com>` so it’s
```
nixjdm: The changes made by black and the import orderer (whichever one is chosen) would be good to put in this commit, and then added to the ignore list. If you add flake8 or another linter or linting rule, those should probably be kept out since they should require a human to implement. flake8 has a lot of good complaints that won't be fixed by black, by the way.

nixjdm: flake8 should also probably not be added to pre-commit or jenkins tests until it's complaints are handled, to not bring everything else to a halt. Making flake8 happy will require some effort.

black.md
s0undt3ch: Reorder python imports does not account for PyLint comments, inline or before the import, it just leaves them behind or it would also be my choice.

But yes, I agree with you, one import per line to reduce conflicts.

black.md
s0undt3ch: Salt doesn't follow the full pep8 "rules" and the parts it does follow, PyLint handles them, however, once black is enabled, all formatting checks in PyLint will be disabled since black will be responsible for that part.

nixjdm: Yes, I'm aware of that. [Link for those who have seen it.](https://docs.saltstack.com/en/latest/topics/development/conventions/style.html) I don't have an issue with not following pep8 rules strictly. Any divergence that we want to keep could be ignored with a flake8 ignore list.

Apart from style, flake8 does catch some useful things that could be actual bugs, especially with [bugbear](https://github.com/PyCQA/flake8-bugbear). For example (without bugbear), if I see a redefinition of a function, it would be helpful to see an inline ignore and/or comment if it's intentional. On a pre-commit, there'd be a good chance a redefinition is a bug.

black.md
nixjdm: Good point, thanks.

black.md
s0undt3ch: Flake8 would be counter the *black* choice. This an `or` scenario, not an `and` scenario. By this I mean, flake8 hasn't been considered an option and so far, me personally, I don't see us wanting to adopt it. Again, this is my personal view on this matter.

An interesting reading about this subject https://www.kevinpeters.net/auto-formatters-for-python

PyLint does catch the particular example you mentioned and of course others.

nixjdm: I'm not sure why it would be counter to black. Black even uses flake8. Flask also uses both, as an example. The only conflict in their default use I'm aware of is line length, which can be set. Black does make enforcing several rules obsolete, though.

You're right, I should have double checked for that particular overlap. I know there's a lot of overlap - but I don't think it is complete, again especially with bugbear.

I picked that example because I quickly found examples of it in salt, though I think some are just not picked up as ignored like they would be for pylint - different syntax. At the bottom of the list of `flake8 salt | grep F811`, for instance, was the first I checked, and a real example: there's a real catch in salt/utils/virtualbox.py (two imports of `six` before the first is used).

The crux of my wanting flake8 I suppose is thinking that the overlap is not complete, which I can dig into more with some time if you like. I take it as a wink and a nudge (but not proof) that flake8 has many current complaints in salt. I also don't really see why you wouldn't use it. It doesn't seem counter to black or pylint to me, just an addition.

black.md
waynew: I think that's fair enough. While `black` helps when you've got a lot of imports, simply forcing them to one line per import should be enough.

And given the automated nature of the changes it shouldn't really be a big deal - write your imports how you will, and they'll get updated by the tools 👍

black.md
waynew: @nixjdm - my only caution against adding the kitchen sink is that every addition has *a* cost.

In the case of black & pylint, the benefit is absolutely worth the cost.

Is the additional benefit of flake8 worth the cost? I don't know. I think if there were concrete examples of, "oh yeah, this (recent) problem would have been caught by flake8", and the cost for running against the codebase is imperceptible, I'd be +1 for it.

Alternatively, if it adds some several extra seconds, but it still finds other issues I'd be +1 on adding it to our CI pipeline (and perhaps make it optional for people to run locally? I'm not sure about the mechanics of that suggestion). Based on the assumption that *most* of our PRs wouldn't require flake8 changes, of course.

I guess ultimately - I don't think we need to explicitly add flake8 here, but I'm not dead set against it.

nixjdm: @waynew Oh I agree about the speed, and I can demonstrate value better.

I ran black twice and timed it; once to fix things (pretty slow) and a second time on a the fixed code. The second time took ~1s. I was assuming roughly the same would hold for flake8, but I can't trivially fix those complaints to test it. If it took much more time than black, I wouldn't want to add it to pre-commit either.

You got me thinking though. flake8 has the ability to parse a `git diff` / `git diff --cached` to only lint the files that were touched, with `git diff | flake8 --diff`. I imagine the time that took on pre-commit would be imperceptible. Since this is such a large repo, if something similar can be done for black and pylint too, that could speed things up noticeably. Regardless though, flake8 shouldn't be added to pre-commit unless/until it's complaints are ignored or fixed.

Running flake8, there are many things found now that I'd be suspicious of, that it would prevent, and I quickly found a couple actual bugs, and lots of things to be suspicious of.

For a few examples:

- Running `flake8 salt/modules | rg "undefined name" | rg -v "__"` (`grep` would work too - filtering out the dunders):
    - [Error formatting with `exc` that wasn't defined](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/modules/cp.py#L147-L148)
    - [The same mistake](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/modules/file.py#L2420-L2425)
- The opposite, local variables that aren't used. These could very well could be bugs, but it's less obvious since you have to understand what it desired. `flake8 salt/modules/ | rg "local variable"` finds unused vars like `return_content`, `err`, (and its variants `exc`, `ex`, and `e`), `ret`, and `ret_dict` that all seem pretty suspect to me.
- `flake8 salt/modules/ | rg "redef"` finds:
    - Simple mistakes like [importing `logging` twice](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/modules/status.py#L12-L17), or [`os` twice](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/modules/ebuildpkg.py#L19-L44) which aren't a bugs (but should be cleaned)
    - [Overwriting an import name with a var](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/modules/win_file.py#L1419) (dirname is imported earlier), which is more suspicious,
    - [Importing `six` in two different ways](https://github.com/saltstack/salt/blob/e37fc20e7cf552425a7e9ba3bfb81e2b2722ed23/salt/utils/virtualbox.py#L20-L26), which is either a bug, or misleading the developer with the first import.

Hopefully that better demonstrates it's value. It finds many more things I'm not listing, of course. The speed is a real concern, but if it isn't fast enough after fixing things, `git diff | flake8 --diff` ought to solve that problem.

nixjdm: It looks like some of this could be achieved by trimming the [Salt's pylint ignore list](https://github.com/saltstack/salt/blob/c746267786cecb5b03467d35d1a5ca24c8bc461a/.testing.pylintrc#L99-L158), which is probably more extensive than it should be, IMO.

Trying to manually compare [pylint's rules](https://pylint.readthedocs.io/en/latest/technical_reference/features.html) to [flake8's](http://flake8.pycqa.org/en/latest/user/error-codes.html) and [bugbear's](https://github.com/PyCQA/flake8-bugbear), I actually haven't spotted rules in pylint not in flake8 yet, though comparing this is pretty difficult. Bugbear, that lake8 extension, has some unique rules though. I'm aware of [salt-pylint](https://github.com/saltstack/salt-pylint) too, but I couldn't quickly find a rules list for it.

Sorry if this feels like I'm beating a (half-)dead horse on this issue. I only mean to help :)

waynew: Makes me think that it would be a good idea to fix a lot of those things... but I think I'll say that flake8 is out of scope for *this* SEP - but it would probably be a good idea to propose a separate SEP to introduce flake8. Or adjust the pylint rules 😝

waynew:
It's time for this SEP to enter the Final Comment period. To sum up the discussion so far:

- Tom can get over the single/double quotes issue based on the number of problems that black will solve
- We've decided we should use single-line mode for isort
- There was discussion around linting/flake8, which seems like a good idea but it should probably be out of scope for this SEP, as it probably needs some alternate discussion, and should probably be its own SEP.

thatch45:
For the record, if there is a reasonable way where we can preserve the single quotes, I would like to see that happen, but I am good with the move otherwise.

waynew:
With 5 votes, this SEP has been accepted.

jfindlay:
I don't know if this is the right place to comment, but there is no documentation on how to test the black linter locally or what to do if your [pull request](https://github.com/saltstack/salt/pull/57850) fails the black linter.  The inline [doc search](https://cse.google.com/cse?cx=011515552685726825874:v1had6i279q&q=black) only yields one indirect reference to black.

I consider myself adequately resourceful and can probably figure it for myself, but it's surprising that such a major change to the development process was not accompanied with a [developer doc](https://docs.saltstack.com/en/master/topics/development/index.html) addition.

s0undt3ch:
Guess you're right. There's no specific documentation with regards to black or how to run it.
We automate that to users/CI using [`pre-commit`](https://docs.saltstack.com/en/latest/topics/development/contributing.html#pre-commit).
`pre-commit run -a black` is what you're looking for.

jfindlay:
Thanks @s0undt3ch.  It might be worth adding to that doc that to get `pre-commit` working, you need to:
- `[sudo] pip install pre-commit` and then
- `cd $my_salt_checkout ; pre-commit install`.

PR #23, opened 2019-10-24
SEP 17 - Job acknowledgement events

lukasraska:
Rendered markdown: https://github.com/lukasraska/salt-enhancement-proposals/blob/job-ack-event/job-ack-event.md

Basic idea is to be able to determine whether minion received job event, or not (due to various reasons). Particularly usable in integrated systems to determine if retrying is an option.

### DIFF
--- a/job-ack-event.md
+++ b/job-ack-event.md
@@ -1,0 +1,84 @@
+- Feature Name: job-ack-event
+- Start Date: 2019-10-24
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Trigger job acknowledge events from minion, before it starts executing the job.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, there seems to be no possibilty to detect if Minion successfully obtained job event. Although there is `saltutil.running`, which could be used to determine it, it creates the same problem as it's another job execution.
+
+The main focus is for integrations to determine whether the job has been started on minion, or if it went stale (e.g. minion is not executing the job although it should), so it can be safely retried.
+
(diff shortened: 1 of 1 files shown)


thatch45:
I LOVE this SEP, I made the saltutil.running system before beacons and the event bus was in full swing! I would also consider leveraging a beacon that also just fired what jobs are running on a regular cadence which would be much smoother than how we track running jobs now

lukasraska:
Beacon could be an option here, to periodically send all running jobs. That will however not help 100% here. It will almost certainly miss short-running jobs (but then you will receive return event from the job, so not that big of an issue), but it won't be "realtime" and will always fire. 

If the beacon is set up for every minute and the job lasts 30 minutes, you get 30 events instead of single ack. But as temporary solution it's usable. I was also looking into how engines are working on minion, but it seems it can only listen for fired events from minion, not the opposite way (so essentially it works as on master, you can catch what minion sends, not what master sends).

waynew:
Yeah, I don't have much to add other than this is a fantastic idea :)

lukasraska:
@waynew Could we get this moved to "Final Comment Period"? :)

waynew:
@lukasraska yeah, sorry for the delay - there has been much travel and holiday time. I'll bring it up in our core meeting tomorrow.

waynew:
@lukasraska I brought it up in our meeting, and the consensus is 👍

There should be some official votes coming in soon and we can get this approved/merged. In our discussion it came up that this is an awesome idea, but it's potentially a large undertaking. @thatch45 has a bunch of information that he plans to share, but this week is super busy for him. You should be able to expect a response from him next Friday.

Let us know if you have any other questions!

lukasraska:
Great, I have PR for this ready (with tests, more or less), so I will just wait for follow-up

lukasraska:
Any comment to this SEP, @thatch45 ?
If not, I will prepare the implementation and it can be incrementally enhanced as the details arise. Thanks

waynew:
@lukasraska This SEP has been accepted 👍

PR #25, opened 2020-02-27
SEP 19: Run salt-master as non-root user (unprivileged)

harmv:
as suggested in https://github.com/saltstack/salt/issues/55886  created a SEP for this issue.

### DIFF
--- a/0019-master-non-root.md
+++ b/0019-master-non-root.md
@@ -1,7 +1,7 @@
 - Feature Name: [security] run salt-master non-root by default
 - Start Date: 2020-02-28
 - SEP Status: Draft
-- SEP PR: (leave this empty)
+- SEP PR: 25
 - Salt Issue: (leave this empty)
 
 # Summary
--- a/master-non-root.md
+++ b/master-non-root.md
@@ -1,0 +1,58 @@
+- Feature Name: [security] run salt-master non-root by default
+- Start Date: 2020-02-28
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Running any server application, typically accessible via the internet, as root user is a security risk.
+Bugs _do_ happen, in any application.   Apps should run with as little rights as possible.
+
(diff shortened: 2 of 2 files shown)


waynew:
This is SEP 19, feel free to update your PR 👍

thatch45:
I completely agree with you that this should be fixed. But this SEP does not present a viable solution or a way to solve this issue. A SEP needs to evaluate what the blockers are to implement and present a path forward on how to resolve it.

This SEP is also abusive in the language used and should be presented in a more professional way.

thatch45:
The main blocker right now is the configuration loading system, it statically looks for files that are only available to root. We should evaluate how that can either be fixed or if Salt's configuration loading system should be extracted from Salt or replaced so that we can create a re-usable system to bypass these issues.

The POP configuration system is designed to fix this problem and can be a source of consideration or could be enhanced so that it could be used and the configuration loading code could be simplified.

waynew:
@harmv are you planning to update the text of your SEP with the feedback from Tom?

harmv:
> @harmv are you planning to update the text of your SEP with the feedback from Tom?

I've updated the SEP with both remarks from Tom.

waynew:
@harmv Thanks for updating this SEP. We've had some internal discussion on this and we absolutely support the fundamental idea (being able to run salt-master as non-root) - but there are a number of questions that are unasked and unanswered by the SEP in it's current form.

If you are willing to drive this SEP to completion, we'd like to have at least a few discussions about the broader implications behind this. We're planning to at least discuss it in a Community Open Hour, though we're also expecting to have a few more discussions. The change is definitely non-trivial. Is this something that you're interested in, or would you like to have someone else take over?

The-Loeki:
As longtime salt user I'd love for this SEP to 'pass' ;) so I took a stab at furthering the discussion a bit.
Shall I submit a new PR for https://github.com/The-Loeki/salt-enhancement-proposals/blob/SEP-19/0019-master-non-root.md ?

waynew:
@The-Loeki yeah, I think if you're willing to take on the reins, that would be 👍

waynew:
@The-Loeki when you open your PR, if @harmv doesn't have any objections, you can just open it as SEP 19 again. For now I'll go ahead and close this PR.

PR #26, opened 2020-04-02
SEP 20: Add python version support SEP

Ch3LL:
SEP to update python version support policy for Salt.


0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+Fedora releases a new version every 6 months and supports these releases for approximately 13 months
+. Our current support mirrors that of the upstream and they are currently supporting Fedora 30, 31 and
+32. They do not have current EOL dates for these releases yet here:
+https://endoflife.software/operating-systems/linux/fedora
+
+|     OS     |    End of Life              |  Default Python |
+|------------|-----------------------------|-----------------|
+|     30     |    TBD (possibly june 2020) |      3.7        |
+|     31     |    TBD (possibly oct 2020)  |      3.7        |
+|     32     |    TBD (possibly apr 2020)  |      3.8        |
```
myii, line 118: may 2021

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+|     Debian 9      |    June 2022     |  2.7, 3.5  |
+|     Debian 10      |    June 2024    |  2.7, 3.7  |
+
+Based on this chart above all of these Operating Systems will or already have conflicted with
+the Python's End of Life Support dates. We will need to include Python in our packages for these
+Operating Systems especially at any given time when their default Python version is considered
+End of Life. This does impact users who would want to pip install Salt. They will need to install
+a newer version of Python before they can install Salt.
+
+Other Reasonable-Effor Support Linux
```
myii, line 153: Effort

myii:
Thanks, a tough issue covered well.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+detail how we will support Python versions going forward. This policy change is to align Salt Releases
+with Python's End of Life schedule.
+
+# Motivation
+[motivation]: #motivation
+
+Documentation and Communication
+-------------------------------
+One of the motivating factors around this policy change is to make it more clear what our Python
+version support is for each Salt releases. If it is clearly defined, then packagers, community members,
```
twangboy, line 20: ... is to clarify our Python version support for each Salt release.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+Python version may drop in that Salt version. This will need to be clearly communicated beforehand
+with the community.
+
+How will this impact the Sodium release?
+----------------------------------------
+Because Python 3.4 became End Of Life on 2019-03-18, we will only support Python versions >= 3.5
+
+How does this impact all supported Operating Systems
+----------------------------------------------------
+The details below will include the impact for our supported operating systmes included here
```
twangboy, line 69: systems

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+AIX/Solaris
+-----------
+AIX/Solaris are Enterprise packages and will use salt-bin for their packaging mechanism,
+so they will include the correct Python version.
+
+Windows
+-------
+Python does not come pre-installed on Windows operating systems. We have always included
+the Python binary in the package and will include to do that going forward. If someone
```
twangboy, line 82: ... and will continue to do that ...

twangboy: Though, it is my understanding that we will be using pop-build even on Windows, though the effect is still the same, that we are bundling python

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+wants to pip install Salt onto windows they will need to install a supported Python version
+first, just as they always have. In the case for the Sodium release this will mean they
+will need to install Python >=3.5
+
+OSX
+---
+Apple does not clearly state when their operating systems become End of Life, but they do
+always support the three latest versions of OSX. We can estimate these End of Life on some of the
+operating systems based on when the latest ones released. The dates included in the table
```
twangboy, line 91: ... latest one is released ...

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+below are not confirmed by Apple and are estimates to help determine how this policy will
+affect macosx.
```
twangboy, line 93: Mac OSX.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+We have always included the Python binary with the mac packages, so that will not change for
+users that use the mac Python packages. Starting with the Sodium release we will no longer
+support Python 2.7, so if users install Salt on OSX using pip or any other means they will
+need to install Python >= 3.5.
+
+Fedora
+------
+Fedora releases a new version every 6 months and supports these releases for approximately 13 months
+. Our current support mirrors that of the upstream and they are currently supporting Fedora 30, 31 and
```
twangboy, line 110: Is there a space before the period here?

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+|     Rhel 7 |    30 Jun 2024      |      2.7       |
+|     Rhel 8 |    05 May 2029      |      None      |
+|     Ubuntu 16.04  |    April 2021    |    2.7/3.5 |
+|     Ubuntu 18.04  |    April 2023    |    3.6     |
+|     Ubuntu 20.04  |    2030          |    3.8     |
+|     Debian 8      |    30 June 2020  |  2.7, 3.4  |
+|     Debian 9      |    June 2022     |  2.7, 3.5  |
+|     Debian 10      |    June 2024    |  2.7, 3.7  |
+
+Based on this chart above all of these Operating Systems will or already have conflicted with
```
twangboy, line 147: Based on the chart above...

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+the Python's End of Life Support dates. We will need to include Python in our packages for these
```
twangboy, line 148: ... have conflicted with Python's End of ...

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+and the requirements defined in our requirements directory. Once these were installed I ran
+this tool https://pypi.org/project/checkmyreqs/ against all installed Python libraries. This brought to light
+the following libraries that will not support Python 3, which will impact some Salt components:
+
+ - pycassa (salt/modules/cassandra.py, salt/returners/cassandra_return.py)
+ - vbox (salt/modules/vbox_guest.py, salt/modules/vbox_guest.py,
+         salt/modules/vboxmanage.py, salt/utils/virtualbox.py,
+         salt/cloud/clouds/virtualbox.py)
+
+Both of these projects have not had commits recently added to the project in years. This will need to be documented
```
twangboy, line 191: have not had commits added to the project in years.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+of the lowest supported Python version. These contributors can look at Python's Release Notes for
+details around what is added into a release or removed. For example for Python 3.6
+https://docs.Python.org/3/whatsnew/3.6.html outlines some of the changes and removals added to this version.
+
+Salt User
+---------
+If a Salt user is using the packages from https://repo.saltstack.com they will not be affected
+as the Python binary will be included in the package. If they are pip installing the package and
+their operating system Python version is an End of Life Python version they will need to upgrade
+they're Python version before being able to pip install Salt.
```
twangboy, line 215: their

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+Users creating their own packages
+---------------------------------
+If a user creates their own packages they will only be impacted if they are packaging for an operating
+system that includes an End of Life Python Version. They would need to either create their packages using
+salt-bin or include a supported version of Python in their packages.
+
+When will this happen
```
twangboy, line 223: Do we need a question mark here?

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+---------------------
+This Python Version Support change will go into affect as soon as its approved, but will first impact
```
twangboy, line 225: ...as soon as it is approved...

twangboy: or `it's`

0000-py-version-support.md
max-arnold: s/Linux/Operating Systems/

(FreeBSD and SmartOS aren't Linuxes)

sblaisot:
As a salt user, I would expect a PEP named "Python Version Support" to not only cover python version deprecation and how salt stop supporting those versions but also what's saltstack goal in supporting newer python versions and how the core team is organized to reach this goal.

This is especially important when dealing with new OS/distribution versions that come with new python version packaged like we see at the moment with python 3.8 and the capacity of salt to handle this new OS/distribution versions.

0000-py-version-support.md
OrangeDog: Actually, it's macOS now.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+for that operating system.
+
+Including Python in Packages
+----------------------------
+Another motivating factor to make this policy change is our shift to including Python in our
+packages. For the Sodium release we will be using salt-bin (https://github.com/saltstack/salt-bin) to
+package for our Fully Supported Platforms.
+(http://get.saltstack.com/rs/304-PHQ-615/images/SaltStack-Supported-Operating-Systems.pdf)
+This will allow us to easily continue supporting operating systems that include End of Life Python
+versions by default by providing the Python binary in the Salt package included in a Salt release.
```
OrangeDog, line 37: What about when the OS doesn't have an old python? I want to use that, not bloat salt with its own version.
If the distribution doesn't provide its own newer python repo, it would be better to at least package python separately (or use official python repos) than to include it in the salt package.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+The design goal of this SEP is to define a clear policy around Salt's Python version support.
+This new policy defines only supporting major Python versions until they reach End of Life. You can see
+the current status of supported Python branches here https://devguide.Python.org/#status-of-Python-branches
+with the associated End Of Life date for each branch. You can also see details about Python versions that
+have already become End Of Life here: https://devguide.Python.org/devcycle/#end-of-life-branches.
+
+What if a Python version drops support relatively close to a Salt Release?
+--------------------------------------------------------------------------
+We may drop support for a Python version if the End of Life date lands relatively close to a Salt release.
```
aplanas, line 58: Dropping support for a Python version means that developers can now use new features available since the oldest supported version. If the release is close to a EOL of a Python version, there will be no time to add new features (and do mostly QA)

So I wonder if makes sense to still support this Python version, or communicate the drop very early on the release process.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+|     30     |    TBD (possibly june 2020) |      3.7        |
+|     31     |    TBD (possibly oct 2020)  |      3.7        |
+|     32     |    TBD (possibly apr 2020)  |      3.8        |
+
+Based off of these dates and Python's End of Life dates, Fedora will not
+be of any concern with this change in policy. Fedora releases fairly regularly
+and includes the latest Python version. Both packages and pip installs
+will work against supported Fedora versions going forward.
+
+SLES/OpenSuse
```
aplanas, line 125: openSUSE

OrangeDog:
https://github.com/saltstack/salt-enhancement-proposals/pull/26#issuecomment-608255709
Especially as there are still states that don't yet work on py3.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+
+What about dependencies?
+------------------------
+Before dropping an End of Life Python version we should have an idea beforehand if Salt's dependencies
+do not support a particular Python version with our testing suite. Currently our tests use the dependencies
+in the requirements directory in the main Salt project. This does of course mean that if we are not testing
+a module we would not have this information beforehand. We are now requiring tests to be added
+for PRs so this will help ensure this is caught for those modules, but there are still some modules that do
+not have tests currently.
```
aplanas, line 179: should this not be addressed when a change of library version in the requirements.txt goes to the CI under the minimal Python version supported?

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+dependency is updated to the appropriate Python version.
+
+## Who is Affected
+
+Salt Contributors
+-----------------
+When a Python Version is dropped any contributions will need to ensure they align with the syntax
+of the lowest supported Python version. These contributors can look at Python's Release Notes for
+details around what is added into a release or removed. For example for Python 3.6
+https://docs.Python.org/3/whatsnew/3.6.html outlines some of the changes and removals added to this version.
```
aplanas, line 208: Good point. If a module in salt does not work on the minimal supported version, do we have a policy in place to drop those modules?

OrangeDog: Until Python 4, there shouldn't be any breaking changes. Even if there was, the module should be fixed, not dropped.

0000-py-version-support.md
```diff
@@ -0,0 +1,244 @@
...
+How does this impact all supported Operating Systems
+----------------------------------------------------
+The details below will include the impact for our supported operating systmes included here
+and some future operating systems we will be supporting:
+http://get.saltstack.com/rs/304-PHQ-615/images/SaltStack-Supported-Operating-Systems.pdf
+
+
+AIX/Solaris
+-----------
+AIX/Solaris are Enterprise packages and will use salt-bin for their packaging mechanism,
```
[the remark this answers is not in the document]
waynew: Yes, @dmurphy18 has had great success using salt--bin. I don't think I've ever seen him so happy 😂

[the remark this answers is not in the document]
sjorge: Is @dmurphy18 doing anything specific to get zmq to compile? I gave salt-bin a whirl on smartos and omnios and it falls back to zmq==bundeled and then choces on missing linux headers.

0000-py-version-support.md
Ch3LL: yes, i wanted to leave the details for salt-bin/pop-build for a different SEP as this is just focused on python version support. The important thing to note here is that Python is included.

0000-py-version-support.md
Ch3LL: I didn't want to focus on the details around the shift/change to using salt-bin for Salt as there will be a separate SEP for this and this SEP focuses on what versions of python we will support. But yes we will be shifting to including python as a part of the package and don't intend on separating it out. There will be more details in that SEP and I'm not currently assigned to working on that team that is doing the research and work around it currently, so that SEP will have more details than i can provide from the team working on it. I would suggest building your own packages or using pip if this is the approach you want to take for one of the operating systems we package for.

0000-py-version-support.md
Ch3LL: Yeah it will need to be clearly communicated and decided beforehand what release it will be dropped in particularly for contributors. We will need to communicate the deprecation decision at the beginning of a feature release development cycle.

Ch3LL: i'll add a clarification note here around that.

0000-py-version-support.md
OrangeDog: > don't intend on separating it out

If you don't intend to change the plan, why go through the SEP process?

Ch3LL: Apologies, didn't mean to make it sound like I wasn't willing to adjust, rather that this conversation should take place in the other SEP. That SEP will include more information around the actual implementation. The individuals working on that project are still working on research and implementation details they need to work out before a SEP with thorough details can be written. The important thing to note is a supported python version will be included to install, but the details around implementation will be included in a future SEP.

Ch3LL: @thatch45, @cro or @dmurphy18  can  elaborate here though as they are more aware of the implementation details here. My understanding is the plan is to include python in the package, but not sure if there is a way to not include it if a user so chooses.

0000-py-version-support.md
Ch3LL: We do not have all libraries for all salt modules in the requirements file, only ones that have test written for them. And just because a library would install on a particular python version does not mean it fully supports that version. Thats why i had to do the extra step to check the extra libraries by installing the requisites to check version support.

Ch3LL:
@sblaisot good point. I'll add a note that we should check for both python version's being removed or added at the beginning of a development cycle, so we can ensure we add it. To clarify around python 3.8 that will be added in Sodium.

0000-py-version-support.md
thatch45: The official packages that SaltStack will ship will include Python, this solves so many issues that go well beyond just targetting the specific python version, including making sure that deps don't break under the release.
The official packages distributed by SaltStack will never be the ONLY way to install Salt, we will still push Salt to pypi and we will still make all sources and build tools available following open source guidelines. If you want to build and distribute system packages on your own, that follow different guidelines, we will strongly encourage it.
But the burden of packaging Salt for all distributions and situations has grown into such a significant burden that it slows testing, slows development, damages releases and introduces more bugs. The benefits we are gaining by following this process are amazingly clear.

Again, this change does not stop anyone from packaging Salt in a classical manner, but it does allow us to make the install and use of Salt much easier, for users, maintainers, distributors and more.

Finally, to answer the level of added bloat, the python distribution shipping with pop-build binaries, build using the configs in the salt-bin repo, is tiny. The python component ranges from just 7-15 megs depending on the platform. I am unaware of any remotely modern computing system, that runs a salt minion, IOT devices included, where this is not a negligible hit to the total storage space on the device.

dmurphy18: The plan is to include Python 3.7 as a minimum in salt-bin, since this reduces the number of dependencies that have to be managed and also deal with a known quantity. Python 3.7 is picked due to PEP 538 and 540 which issues with locales on some platforms and containers only providing for locale 'C' and POSIX.  Some of those platforms have Python 3.6 and PEP 538 is backported but ran into issues with traceback stackframe being output and other logging issues with exceptions resulting in OOM Killer's being called, which were completely resolved using Python 3.7.  As platforms have new versions of Python, for example: Ubuntu 20.04 and Python 3.8, then salt-bin will leverage Python 3.8.  Salt-bin leverages PyInstaller and picking a self-contained Python version allows for consistency across revisions to a platform.  One only needs to review the changes to Redhat 7 last summer with the changes between Python 3.4 to Python 3.6 and the ensuing upsets in EPEL and rpm naming, to isolate from those issues with Redhat or other platforms.

0000-py-version-support.md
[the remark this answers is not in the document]
dmurphy18: Have not built salt-bin on AIX or Solaris yet, but PyInstaller states it functions on it. But Solaris 10 is fast becoming a candidate, with trying to get an old version of libsodium to build, it used to but not anymore. But @waynew is correct, once Python 3.7 is available (even if you have to put it there on the build machine), building with salt-bin has been a charm, however have found that using --onedir option is the best way to proceed. Also see my comment above on PEP 538 and 540 for locales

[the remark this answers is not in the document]
thatch45: Generally pyzmq will bundle zmq, in all of our tests on Linux, Windows and Mac it has worked just fine.
But you can explicitly bundle compiled libs using pop-build to solve this problem. Take a look at this doc:
https://pop-build.readthedocs.io/en/latest/topics/quickstart.html#using-the-build-addon

0000-py-version-support.md
waynew: Even in Python4 there's unlikely to be breaking changes :upside_down_face:

waynew:
This is SEP 20 - feel free to update your files :)

Ch3LL:
Can i get re-review here or any further concerns? @twangboy @max-arnold @myii @aplanas @OrangeDog @sblaisot

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+
+Windows
+-------
+Python does not come pre-installed on Windows operating systems. We have always included
+the Python binary in the package and will continue to do that going forward. If someone
+wants to pip install Salt onto windows they will need to install a supported Python version
+first, just as they always have. In the case for the Sodium release this will mean they
+will need to install Python >=3.5
+
+OSX
```
OrangeDog, line 94: It's called macOS now. Multiple occurrences.

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+
+Users creating their own packages
+---------------------------------
+If a user creates their own packages they will only be impacted if they are packaging for an operating
+system that includes an End of Life Python Version. They would need to either create their packages using
+salt-bin or include a supported version of Python in their packages.
+
+When will this happen?
+----------------------
+This Python Version Support change will go into affect as soon as it is approved, but will first impact
```
OrangeDog, line 232: effect

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+dependency is updated to the appropriate Python version.
+
+## Who is Affected
+
+Salt Contributors
+-----------------
+When a Python Version is dropped any contributions will need to ensure they align with the syntax
+of the lowest supported Python version. These contributors can look at Python's Release Notes for
+details around what is added into a release or removed. For example for Python 3.6
+https://docs.Python.org/3/whatsnew/3.6.html outlines some of the changes and removals added to this version.
```
OrangeDog, line 215: > ensure they work with the lowest supported Python version. These contributors can look at Python's Release Notes for details of what has been removed or deprecated. For example https://docs.Python.org/3/whatsnew/3.6.html#removed lists what was removed in Python 3.6.

Most changes are not syntax. No "alignment" is required, and new features aren't relevant.

I notice that this fails to account for existing code that might break. Unless you require all contributors to maintain their contributions indefinitely?

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+
+Salt User
+---------
+If a Salt user is using the packages from https://repo.saltstack.com they will not be affected
+as the Python binary will be included in the package. If they are pip installing the package and
+their operating system Python version is an End of Life Python version they will need to upgrade
+their Python version before being able to pip install Salt.
```
OrangeDog, line 222: > operating system Python is not supported by the version of salt being installed

Whether it's EoL now doesn't matter if it wasn't when salt was released.

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+
+When will this happen?
+----------------------
+This Python Version Support change will go into affect as soon as it is approved, but will first impact
+the Sodium Release. The Sodium Release is due to be released in June and will support Python versions >= 3.5.
+
+## Alternatives
+[alternatives]: #alternatives
+
+- Alternative is to keep supporting insecure Python versions that line up with supported operating systems
```
OrangeDog, line 238: Supported operating systems will not have an insecure Python. The work that RedHat/Debian/etc. does is to ensure their LTS systems continue to have necessary security patches, even if the upstream developers don't provide them.

0020-py-version-support.md
OrangeDog: Are new contributions still allowed to use deprecated features? What's the timeline for updating new deprecations that arise in existing code? Will there be an automated system to capture them (e.g. from CI/integration tests)?

0020-py-version-support.md
```diff
@@ -0,0 +1,251 @@
...
+-----------
+
+|     OS     |    End of Life      | Default Python |
+|------------|---------------------|----------------|
+|     Rhel 6 |    30 Nov 2020      |      2.6       |
+|     Rhel 7 |    30 Jun 2024      |      2.7       |
+|     Rhel 8 |    05 May 2029      |      None      |
+|     Ubuntu 16.04  |    April 2021    |    2.7/3.5 |
+|     Ubuntu 18.04  |    April 2023    |    3.6     |
+|     Ubuntu 20.04  |    2030          |    3.8     |
```
myii, line 149:
```suggestion
|     Ubuntu 20.04  |    April 2025    |    3.8     |
```

Source: https://ubuntu.com/about/release-cycle.

0020-py-version-support.md
Ch3LL: I can just remove the word "syntax" and ensure it states contributions need to align with the lowest supported python version. And yes a contributor would need to align with both removals and additions.

New features are relevant. For example, if we begin to support >=Python 3.6 then a contributor can now use fstrings. Whereas when we support  >=python3.5 we cannot accept contributions with fstrings.

You cannot use deprecated features for your contributions. For example if we start supporting >=Python 3.6 you cannot use this deprecated code anymore:

`traceback.Ignore class and traceback.usage, traceback.modname, traceback.fullmodname, traceback.find_lines_from_code, traceback.find_lines, traceback.find_strings, traceback.find_executable_lines methods were removed from the traceback module. They were undocumented methods deprecated since Python 3.2 and equivalent functionality is available from private methods.`

For updating deprecations/testing I specifically call out when we add new python versions here:

https://github.com/saltstack/salt-enhancement-proposals/pull/26/files#diff-d745c39ab3665fe44caa19ff2a6dffdaR62

So at the beginning of each feature development cycle we need to see what python versions are being not only dropped but added and ensure testing is added around that version which should capture the deprecations.

0020-py-version-support.md
Ch3LL: true. i can clean that up to be more clear

0020-py-version-support.md
Ch3LL: not all of our supported operating systems follow this approach. I can change the syntax to say End of Life instead of insecure though to align with all of them.

0020-py-version-support.md
OrangeDog: There's a big difference between "can use fstrings" and "must use fstrings". To "align with additions" means the latter, which IMO is unreasonable.

0020-py-version-support.md
OrangeDog: Which systems does Salt support, that do not support their own core packages?

0020-py-version-support.md
Ch3LL: yeah i agree with you that its not a must for new features, I can lighten up the language there.

0020-py-version-support.md
Ch3LL: I could not see similar security support for macOS, aix and solaris. Please correct me if im wrong.

OrangeDog: I'd be very surprised if IBM and Oracle are leaving unpatched public vulnerabilities in their supported commercial systems. Looks like Apple have or are shortly going to stop packaging Python themselves.

Ch3LL: Yeah I honestly just couldn't find documentation. Either way I updated the phrase to be End of Life here instead of insecure.

Ch3LL:
@OrangeDog @myii made updates. ready for re-review. thanks

Ch3LL:
any other concerns here? I will put this in final comment period starting next week since we are still waiting on re-review from @OrangeDog and @myii

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+
+Documentation and Communication
+-------------------------------
+One of the motivating factors around this policy change is to clarify our Python version support
+for each Salt release. If it is clearly defined, then packagers, community members, and Salt users will
+know when a Python version will be dropped for a given Salt release. In the past we had never clearly
+defined how our Python version support lined up with Python's End of Life schedule. We previously made a
+best effort to support the Python versions that lined up with the operating systems Python versions, but never clearly
+defined this and did not always follow this in every case. For example, when we dropped Python 2.6 support
+we still supported Redhat 6, which had Python 2.6 installed by default. To continue supporting Redhat 6 while also
```
myii, line 25:
```suggestion
we still supported Red Hat 6, which had Python 2.6 installed by default. To continue supporting Red Hat 6 while also
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+(http://get.saltstack.com/rs/304-PHQ-615/images/SaltStack-Supported-Operating-Systems.pdf)
+This will allow us to easily continue supporting operating systems that include End of Life Python
+versions by default by providing the Python binary in the Salt package included in a Salt release.
+
+Security and Maintenance Concerns
+---------------------------------
+The last motivating factor surrounds security and maintenance concerns. Continuing to support an End
+of Life Python version adds additional maintenance around testing and development for each version.
+Supporting these versions also puts users at a security risk, since no more security fixes will be
+submited and released with that Python version if the operating system does not choose to update
```
myii, line 43:
```suggestion
submitted and released with that Python version if the operating system does not choose to update
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+it themselves.
+
+# Design
+[design]: #detailed-design
+
+The design goal of this SEP is to define a clear policy around Salt's Python version support.
+This new policy defines only supporting major Python versions until they reach End of Life. You can see
+the current status of supported Python branches here https://devguide.Python.org/#status-of-Python-branches
```
myii, line 51:
```suggestion
the current status of supported Python branches here https://devguide.python.org/#status-of-Python-branches
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+Fedora releases a new version every 6 months and supports these releases for approximately 13 months.
+Our current support mirrors that of the upstream and they are currently supporting Fedora 30, 31 and
+32. They do not have current EOL dates for these releases yet here:
+https://endoflife.software/operating-systems/linux/fedora
+
+|     OS     |    End of Life              |  Default Python |
+|------------|-----------------------------|-----------------|
+|     30     |    TBD (possibly june 2020) |      3.7        |
+|     31     |    TBD (possibly oct 2020)  |      3.7        |
+|     32     |    TBD (possibly may 2021)  |      3.8        |
```
myii, line 125:
```suggestion
|     30     |    TBD (possibly June 2020) |      3.7        |
|     31     |    TBD (possibly Oct 2020)  |      3.7        |
|     32     |    TBD (possibly May 2021)  |      3.8        |
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+
+Based off of these dates and Python's End of Life dates, Fedora will not
+be of any concern with this change in policy. Fedora releases fairly regularly
+and includes the latest Python version. Both packages and pip installs
+will work against supported Fedora versions going forward.
+
+SLES/openSUSE
+-------------
+The General End of Life dates do conflict with the current Python Version
+End of Life dates. Since Suse does the packaging here they have been informed
```
myii, line 135:
```suggestion
End of Life dates. Since SUSE does the packaging here they have been informed
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+when these version conflicts occur.
+
+Other Linux
+-----------
+
+|     OS     |    End of Life      | Default Python |
+|------------|---------------------|----------------|
+|     Rhel 6 |    30 Nov 2020      |      2.6       |
+|     Rhel 7 |    30 Jun 2024      |      2.7       |
+|     Rhel 8 |    05 May 2029      |      None      |
```
myii, line 146:
```suggestion
|     RHEL 6 |    30 Nov 2020      |      2.6       |
|     RHEL 7 |    30 Jun 2024      |      2.7       |
|     RHEL 8 |    05 May 2029      |      None      |
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+Operating Systems especially at any given time when their default Python version is considered
+End of Life. This does impact users who would want to pip install Salt. They will need to install
+a newer version of Python before they can install Salt.
+
+Other Reasonable-Effort Supported Operating Systems
+---------------------------------------------------
+
+Arch
+----
+Arch is a rolling release and is consistantly updating their Python version, so this is not
```
myii, line 165:
```suggestion
Arch is a rolling release and is consistently updating their Python version, so this is not
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+What about dependencies?
+------------------------
+Before dropping an End of Life Python version we should have an idea beforehand if Salt's dependencies
+do not support a particular Python version with our testing suite. Currently our tests use the dependencies
+in the requirements directory in the main Salt project. This does of course mean that if we are not testing
+a module we would not have this information beforehand. We are now requiring tests to be added
+for PRs so this will help ensure this is caught for those modules, but there are still some modules that do
+not have tests currently.
+
+For the Sodium Release, I installed the libraries from anything defined with :depends: in the documentation
```
myii, line 188:
```suggestion
For the Sodium Release, I installed the libraries from anything defined with `:depends:` in the documentation
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+Python version. For example, f-strings were added in Python 3.6 so when our Python version support is updated
+to using >= Python 3.6 a contributor will be able to use f-strings now.
+
+A contributor will also need to be aware of what is deprecated and removed in all supported Python releases when
+making new contributions. For example, in the Python 3.8 release the `macpath` module was removed,
+so a contributor would need to ensure they do not write any new contributions using this module.
+
+A contributor can look at Python's Release Notes for details around what is added into a release or removed.
+For example for Python 3.6
+https://docs.Python.org/3/whatsnew/3.6.html outlines some of the changes and removals added to this version.
```
myii, line 222:
```suggestion
https://docs.python.org/3/whatsnew/3.6.html outlines some of the changes and removals added to this version.
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+it themselves.
+
+# Design
+[design]: #detailed-design
+
+The design goal of this SEP is to define a clear policy around Salt's Python version support.
+This new policy defines only supporting major Python versions until they reach End of Life. You can see
+the current status of supported Python branches here https://devguide.Python.org/#status-of-Python-branches
+with the associated End Of Life date for each branch. You can also see details about Python versions that
+have already become End Of Life here: https://devguide.Python.org/devcycle/#end-of-life-branches.
```
myii, line 53:
```suggestion
have already become End Of Life here: https://devguide.python.org/devcycle/#end-of-life-branches.
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+to be added. Tests running against the new Python version will need to be added and run on the
+master branch. Any issues found will need to have issues created and assigned out for the
+scheduled release.
+
+How will this impact the Sodium release?
+----------------------------------------
+Because Python 3.4 became End Of Life on 2019-03-18, we will only support Python versions >= 3.5
+
+How does this impact all supported Operating Systems
+----------------------------------------------------
```
myii, line 75:
```suggestion
How does this impact all supported Operating Systems?
-----------------------------------------------------
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+
+SLES/openSUSE
+-------------
+The General End of Life dates do conflict with the current Python Version
+End of Life dates. Since Suse does the packaging here they have been informed
+of this policy change and will include a resolution around this with their packages
+when these version conflicts occur.
+
+Other Linux
+-----------
```
myii, line 140:
```suggestion
Other Linuxes
-------------
```

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+## Who is Affected
+
+Salt Contributors
+-----------------
+When a Python Version is dropped the contributor can now use new features in the lowest supported
+Python version. For example, f-strings were added in Python 3.6 so when our Python version support is updated
+to using >= Python 3.6 a contributor will be able to use f-strings now.
+
+A contributor will also need to be aware of what is deprecated and removed in all supported Python releases when
+making new contributions. For example, in the Python 3.8 release the `macpath` module was removed,
```
myii, line 217: Using the backticks for `macpath` here (which I'm in favour of) -- however, from a consistency standpoint, the other uses of "macpath" in this document are not using backticks, so I believe it should be one or the other.

Ch3LL:
okay ready for re-review @myii  thanks :)

Ch3LL:
thanks for catching that let me make the H uppercase.

waynew:
This SEP has entered the final comment period, let us know if there's anything else that needs to be addressed!

terminalmage:
I like the level of detail WRT different platforms, but there should be a matrix listed somewhere on the docs which details which platforms have a bundled Python, which Python release will be bundled, etc.

Additionally, how does adding pip packages work for bundled Python? If these aren't "onedir" POP builds (I didn't see this specified), that makes installing pip packages problematic, as the default method Salt uses to decide which pip to use is to run `$python -m pip .....`, where `$python` is `sys.executable`. I'm not sure what `sys.executable` returns for a fully-contained build, but it probably won't work with the pip execution module.

I'm not sure any of these concerns merits changes to the contents of this SEP, but at minimum there should be open issues labeled `Sodium` which ensure that we address them before the release.

waynew:
We can still pip install with the one-dir pop-build distributions, so no issue there.

IIUC the plan is to provide pip installable Salt (for which this SEP will be particularly relevant), and then pop-build packages via repo.saltstack.com, which will bundle Python.

terminalmage:
So, the POP builds will be onedir and not single-binary? That question really wasn't answered. I'm aware of how onedir builds work, the question is an effort to confirm that basic pip functionality will be present in these builds. With so many states and execution modules relying on third-party Python bindings, pip support is pretty important to Salt users.

OrangeDog:
On the subject of bundling then...

Many users and formula use `pkg.installed` to get salt dependencies from their system packages. This will break all of that then?

waynew:
> So, the POP builds will be onedir and not single-binary? That question really wasn't answered.

Both.

IIRC, the single-binary can be installed in a onedir fashion. I could be wrong and pop-build has to produce two different packages, but my understanding is that regardless of the mechanics, it's pretty trivial to get both onedir and single-binary.

> Many users and formulae use pkg.installed to get salt dependencies from their system packages. This will break all of that then?

Are you referring to module dependencies, or salt core dependencies? Salt core dependencies will be included. But for the module dependencies... my gut says that would break **unless** we add the system library to the PYTHONPATH.

I'll double check on that.

OrangeDog:
> that would break unless we add the system library to the PYTHONPATH

Yes, module dependencies. Also, I don't think that would work unless the system and bundled versions are the same, at which point what was the point of bunding it?

I raised all these issues [above](https://github.com/saltstack/salt-enhancement-proposals/pull/26#pullrequestreview-387088838) but that got shut down because it's for a future SEP. It's not even been proposed yet, but it seems like a forgone conclusion you're going to start bundling python, whatever anyone says.

And if it's supposed to start happening in Sodium, it's rather late to be leaving it...

waynew:
Some clarification:
- onedir and single binary are different artifacts
- pip installs work already for onedir. There is work happening for pip installs with the single-binary approach.
- no, system packages module dependencies are not supported. 
> Also, I don't think that would work unless the system and bundled versions are the same?

*only* for compiled packages. If it's a pure python package, and I *think* even if there's an ABI compatibility, but for sure pure python packages would work just fine.

0020-py-version-support.md
```diff
@@ -0,0 +1,258 @@
...
+best effort to support the Python versions that lined up with the operating systems Python versions, but never clearly
+defined this and did not always follow this in every case. For example, when we dropped Python 2.6 support
+we still supported Red Hat 6, which had Python 2.6 installed by default. To continue supporting Red Hat 6 while also
+removing support for the End of Life Python 2.6 version, we included Python 2.7 in our packages
+for that operating system.
+
+Including Python in Packages
+----------------------------
+Another motivating factor to make this policy change is our shift to including Python in our
+packages. For the Sodium release we will be using salt-bin (https://github.com/saltstack/salt-bin) to
```
bryceml, line 32: Salt-bin is now pop-build https://gitlab.com/saltstack/pop/pop-build/

waynew:
```suggestion
packages. For the Sodium release we will be using pop-build (https://gitlab.com/saltstack/pop/pop-build/) to
```

0000-py-version-support.md
waynew:
```suggestion
Based on the chart above all of these Operating Systems will or already have conflicted with
```

waynew:
This SEP has received enough votes to be accepted 🎉

PR #27, opened 2020-04-14
SEP 21: Proposed changes to onchanges state requisite.

garethgreenaway:
Adding SEP for proposed changes to onchanges state requisite.


sblaisot:
Why change the meaning of `onchanges:` instead of introducing an `onchanges_all` for the new meaning?

This way, `onchanges:` would be an alias of `onchanges_any:` like it always was keeping backward compatibility and the new `onchanges_all:` would allow to trigger only if all dependant states have changes.

The proposed change will break any existing state that have multiple `onchanges:` requisites and I can see no obvious reason to make this breaking change and break all these existing states.

OrangeDog:
Salt staff still seem to think that `onchanges` and `onchanges_any` do the same thing.
As [I tried to explain](https://github.com/saltstack/salt/issues/56133#issuecomment-588384040), they do not.
Changing either of them would make it inconsistent with all the other requisites.

sblaisot:
Adding links to documentation:
- [`onchanges`](https://docs.saltstack.com/en/latest/ref/states/requisites.html#onchanges)
- [`onchanges_any`](https://docs.saltstack.com/en/latest/ref/states/requisites.html#onchanges_any)

OrangeDog:
`onchanges` - all states must return True  
`onchanges_any` - only one state must return True

waynew:
This is SEP 21, feel free to update your file :)

garethgreenaway:
@OrangeDog You' are correct, they currently do not provide the same functionality but they are very close.  Currently `onchanges_any` will trigger if any of the watched states returns a result of True and returns changes, this is correct and matches with the other `_any` requisites.  The `onchanges` requisite will return if ALL the states return True and if any of the states return changes.  It is my belief that this functionality is incorrect, `onchanges` should trigger only if all the states are returning changes.

@sblaisot The change to `onchanges` is motivated by the belief that the functionality is currently incorrect and that the documentation indicates a discrepancy between what is documented and the actual functionality.  Any implementation would be done with at least a two release announcement to avoid breaking existing states.

OrangeDog:
@garethgreenaway should `watch` be changed as well then? Currently it triggers the watch function if any of the states have changes, not all of them.

garethgreenaway:
@OrangeDog I would say so, since we have `watch_any`, we should make sure the requisites are consistent.

OrangeDog:
> we should ensure that our requisites are uniform and consistent

As far as I can tell, they currently are. If they are all changed, how would you get back the current behaviour of `watch` and `onchanges` (and `prereq`)? The `_any` versions will ignore failures.

waynew:
@OrangeDog a state can:

- succeed (return True)
- succeed with changes (return True, with `['changes']`)
- fail (return False)
- fail with changes (return False, with `['changes']`)

That means that the scenarios are:

-  all states succeed, with no changes.
- all states succeed, and have changes
- all states succeed, with only some changes
- some fail, some succeed, with no changes
- some fail, some succeed, with changes on success
- some fail with changes, some succeed with no changes
- some fail with changes, some succeed with changes
- all states fail with no changes
- all states fail, some with changes

Right now, IIUC, it's not exactly consistent (at least consistently named) what the reqs and `_any` reqs do across the cases.

OrangeDog:
They can also return `None`, with or without changes, but that's similarly not relevant.

Ignoring `onfail` for now, all plain reqs require all referents to return `True`, and all `_any` reqs require any referent to return `True`. Then `watch`, `onchanges`, and `prereq` have an additional requirement that any of the referents have changes. The two things are unrelated, and consistent.

sagetherage:
Closing as `Withdrawn` as this needs to be overhauled and re-written as per @garethgreenaway

PR #28, opened 2020-05-20
SEP 22: Move old-releases to a different domain name/url

bryceml:
pretty version: https://github.com/bryceml/salt-enhancement-proposals/blob/old_releases/old-releases.md

### DIFF
--- a/old-releases.md
+++ b/old-releases.md
@@ -126,3 +126,8 @@ for comments from May 20-26, 2020. The detailed timeline is below:
 - We would have to update documentation describing the new behavior.
 - It will take effort for someone to move unsupported versions to the archive
   domain name.
+- People pinning to minor releases on repo.saltstack.com may have unexpected
+  404 errors when we move the insecure release to archive.repo.saltstack.com
+  depending on the timeline of transition and whether they are aware of the CVE
+  release.  We recommend pinning to the major or latest release to avoid that
+  issue.
--- a/old-releases.md
+++ b/old-releases.md
@@ -21,9 +21,10 @@ would leave many stranded while trying to upgrade to later versions and is
 therefore not viable either. This leaves us with the approach of moving the
 older, unsupported versions to an archive location. In examining this, we
 assessed other open projects and their approach to older, unsupported releases.
-We found many following the same proposed approach that we have in this SEP. An
-example is Ubuntu, which makes this distinction easy by using two separate
-domain names: old-releases.ubuntu.com and archive.ubuntu.com. We should emulate
+We found many following the same proposed approach that we have in this SEP.
+Examples include Ubuntu and Debian, which make this distinction easy by using
+two separate domain names: old-releases.ubuntu.com and archive.ubuntu.com for
+Ubuntu, and archive.debian.org and deb.debian.org for Debian. We should emulate
 their behavior by moving our old unsupported releases to a separate URL.
 
 On the same note, unsupported PyPi releases should also be moved to a static
@@ -43,7 +44,7 @@ costs for them and SaltStack.
 # Design
 [design]: #detailed-design
 Whenever a major release goes out of CVE support, we will move the contents to
-https://old-releases.repo.saltstack.com, for example:
+https://archive.repo.saltstack.com, for example:
 
 ```
 https://repo.saltstack.com/yum/redhat/7/x86_64/2018.3/
@@ -51,7 +52,7 @@ https://repo.saltstack.com/yum/redhat/7/x86_64/2018.3/
 would move to
 
 ```
-https://old-releases.repo.saltstack.com/yum/redhat/7/x86_64/2018.3/
+https://archive.repo.saltstack.com/yum/redhat/7/x86_64/2018.3/
 ```
 and all of the point releases under
 
@@ -61,12 +62,12 @@ https://repo.saltstack.com/yum/redhat/7/x86_64/archive/
 that are part of that major release would move to
 
 ```
-https://old-releases.repo.saltstack.com/yum/redhat/7/x86_64/archive/
+https://archive.repo.saltstack.com/yum/redhat/7/x86_64/archive/
 ```
 
 If a point release is affected by a CVE that is part of a major supported
 branch, that point release will also be moved to
-https://old-releases.repo.saltstack.com. For example,
+https://archive.repo.saltstack.com. For example,
 
 ```
 https://repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
@@ -74,16 +75,15 @@ https://repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
 was affected by the recent CVE so it would be moved to
 
 ```
-https://old-releases.repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
+https://archive.repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
 ```
 
 This would mean someone using the bootstrap script would have to specify `-R
-old-releases.repo.saltstack.com` if they wanted to install an
-unsupported/insecure release.  Other methods of installation would also require
-updating the URL with the new domain name.
+archive.repo.saltstack.com` if they wanted to install an unsupported/insecure
+release.  Other methods of installation would also require updating the URL
+with the new domain name.
 
-https://old-releases.repo.saltstack.com will be S3 syncable just like the main
-repo.
+https://archive.repo.saltstack.com will be S3 syncable just like the main repo.
 
 We will add a notice to the main page of repo.saltstack.com explaining this
 behavior.
@@ -124,5 +124,5 @@ for comments from May 20-26, 2020. The detailed timeline is below:
   https://repo.saltstack.com/yum/redhat/salt-repo-2015.5-1.el5.noarch.rpm which
   may inconvenience people who would like to use them.
 - We would have to update documentation describing the new behavior.
-- It will take effort for someone to move unsupported versions to the
-  old-releases domain name.
+- It will take effort for someone to move unsupported versions to the archive
+  domain name.
--- a/old-releases.md
+++ b/old-releases.md
@@ -102,9 +102,9 @@ this SEP is urgent and running on an expedited timeline. The SEP will be open
 for comments from May 20-26, 2020. The detailed timeline is below:
 - May 20, 2020: SEP Announcement
 - May 21, 2020: SEP discussion during the Community Open Hour
-- May 26, 2020: SEP closed to comments
-- May 27, 2020: SEP closes
-- June 1, 2020: Unsupported releases moved to separate URL
+- May 29, 2020: SEP closed to comments
+- June 1, 2020: SEP closes
+- June 8, 2020 or later: Unsupported releases moved to separate URL
 
 ## Alternatives
 [alternatives]: #alternatives
--- a/old-releases.md
+++ b/old-releases.md
@@ -1,0 +1,127 @@
+- Feature Name: move non-supported releases to a separate domain name
+- Start Date: 2020-05-20
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+With the recent CVE events, we have observed and have been alerted to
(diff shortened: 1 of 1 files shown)


sblaisot:
Seems to me like a good compromise 👍 

Maybe the 404 error returned when trying to access an unsupported release / moved package could be customized to say something like "this release has been moved to old-releases repo"? Not sure how package manager would deal with that info though

terminalmage:
`old-releases` seems like an odd name, a lot of distros move older releases to `archive.domain.tld`, and this seems like a good convention to follow.

jleroy:
@terminalmage +1

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+make a decision on what to do with older, unsupported releases.
+
+We have three potential paths: release updated packages for older versions,
+remove older versions all together, or move older versions to an archive
+location that is separated from currently supported versions. From a time and
+effort standpoint, the option to update is simply not viable and would set the
+project back considerably. The option to completely remove the older packages
+would leave many stranded while trying to upgrade to later versions and is
+therefore not viable either. This leaves us with the approach of moving the
+older, unsupported version to an archive location. In examining this, we
```
myii, line 22:
```suggestion
older, unsupported versions to an archive location. In examining this, we
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+assessed other open projects and their approach to older, unsupported releases.
+We found many following the same proposed approach that we have in this SEP. An
+example is Ubuntu, which makes this distinction easy by using two seperate
```
myii, line 25:
```suggestion
example is Ubuntu, which makes this distinction easy by using two separate
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+was affected by the recent CVE so it would be moved to
+
+```
+https://old-releases.repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
+```
+
+This would mean someone using the bootstrap script would have to specify `-R
+old-releases.repo.saltstack.com` if they wanted to install an
+unsupported/insecure release.  Other methods of installation would also require
+updating the url with the new domain name.
```
myii, line 83:
```suggestion
updating the URL with the new domain name.
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+
+https://old-releases.repo.saltstack.com will be s3 syncable just like the main repo.
```
myii, line 85:
```suggestion
https://old-releases.repo.saltstack.com will be S3 syncable just like the main repo.
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+
+We will add a notice to the mainpage of repo.saltstack.com explaining this
```
myii, line 87:
```suggestion
We will add a notice to the main page of repo.saltstack.com explaining this
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+behavior.
+
+With regards to the source tarballs hosted on PyPi, we will host them on a web
+server and they will be available on a static HTML page. The reason to do this
```
myii, line 91:
```suggestion
server and they will be made available on a static HTML page. The reason to do this
```

old-releases.md
```diff
@@ -0,0 +1,127 @@
...
+```
+was affected by the recent CVE so it would be moved to
+
+```
+https://old-releases.repo.saltstack.com/yum/redhat/7/x86_64/archive/3000/
+```
+
+This would mean someone using the bootstrap script would have to specify `-R
+old-releases.repo.saltstack.com` if they wanted to install an
+unsupported/insecure release.  Other methods of installation would also require
```
myii, line 82: This is more work but how about adding a warning message to the bootstrap script, when a vulnerable version is being requested for installation?  Goes along with the point made below, "to make it explicit, and not easy, to install older versions."  This may be wider-reaching than just `old-releases.repo.saltstack.com`, though, i.e. enough to check the version number rather than checking the source of the package.

rossengeorgiev:
1. Consider having both retired, and current releases under the new domain.  CI systems, may need to hardcode/template the URL to a specific version. Currently, that is possible with for example: `https://repo.saltstack.com/py3/redhat/$releasever/$basearch/archive/<VERSION>`. The repo RPMs won't work, but I think that is fine. 

   Put another way, I am suggesting to keep the current structure (or similar) under the new domain, and then only have current releases under `repo.saltstack.com`. 

2. Additionally, rework the following menu, by indicating current releases and no longer supported ones. For example, 2019 is previous, but is still supported. 

image

3. On a UX note, consider turning the version drop down menu into a tabs. Similar to how https://packages.ubuntu.com/ does it. All current releases are listed at the top at all times. That saves users a extra click and speeds up interaction, while indicating the status of a release at a glance. Could be color coded, or somehow separated.

bryceml:
> `old-releases` seems like an odd name, a lot of distros move older releases to `archive.domain.tld`, and this seems like a good convention to follow.

@terminalmage I was just trying to follow convention.  I see debian uses `archive.debian.org`, can't seem to find anyone else that does, but I like archive better as well.  What if we used `archive.repo.saltstack.com` ?  I like the idea of having everything related to `repo.saltstack.com` end in `repo.saltstack.com`

terminalmage:
I like that.

jleroy:
> I see debian uses `archive.debian.org`, can't seem to find anyone else that does, but I like archive better as well.

FYI Fedora uses `archives.fedoraproject.org`, Ubuntu uses `archive.ubuntu.com`, RHEL uses `archive.download.redhat.com`

`archive.repo.saltstack.com` looks good to me.

rossengeorgiev:
Ubuntu, definitely uses `old-releases`, see https://releases.ubuntu.com/

jleroy:
You're right, supported releases are hosted on `archive.ubuntu.com`, older releases on `old-releases.ubuntu.com` which is... disturbing.

myii:
A thought that got triggered while watching the Open Hour, specifically the comment by @sblaisot.

Speaking about over a period of time, would there be any useful way to distinguish between old releases that are simply unsupported (e.g. `2019.2.5`) and those that are unsupported AND contain vulnerabilities (e.g. `2019.2.3`)?

bryceml:
> A thought that got triggered while watching the Open Hour, specifically the comment by @sblaisot.
> 
> Speaking about over a period of time, would there be any useful way to distinguish between old releases that are simply unsupported (e.g. `2019.2.5`) and those that are unsupported AND contain vulnerabilities (e.g. `2019.2.3`)?

all point releases of a supported major release will remain except those affected by CVEs, so 

2019.2.5 would remain on repo.saltstack.com

so
```
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.5/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.4/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/2019.2/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/latest/
etc.
```
would all remain and 
```
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/2019.2/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/latest/
```
would be updated to point to the latest respective point releases (they are symlinks), only
```
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.3/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.2/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.1/
https://repo.saltstack.com/py3/ubuntu/18.04/amd64/archive/2019.2.0/
```
would be moved to https://archive.repo.saltstack.com

Hopefully this explains it better.

I imagine we should define a period of time where a release that is moved is available on both domain names, maybe a day or 2 to give people time to move without interuption.

myii:
> 2019.2.5 would remain on repo.saltstack.com

@bryceml I probably didn't state the scenario clearly: I meant in the longer term of things, say when Sodium is released and `2019.2.5` become obsolete (or even by Magnesium).  I understand that it will be moved to `archive.repo.saltstack.com`, right?  If so, would there be a way to distinguish between this and the vulnerable versions?

bryceml:
> > 2019.2.5 would remain on repo.saltstack.com
> 
> @bryceml I probably didn't state the scenario clearly: I meant in the longer term of things, say when Sodium is released and `2019.2.5` become obsolete (or even by Magnesium). I understand that it will be moved to `archive.repo.saltstack.com`, right? If so, would there be a way to distinguish between this and the vulnerable versions?

since 2019.2 would be out of CVE support at that point, we wouldn't take the time to determine what CVE's, if any, it was vulnerable to.  Since it is out of support, that becomes a continually changing unknown we don't have time to constantly investigate for old releases.

sblaisot:
> since 2019.2 would be out of CVE support at that point, we wouldn't take the time to determine what CVE's, if any, it was vulnerable to. Since it is out of support, that becomes a continually changing unknown we don't have time to constantly investigate for old releases.

When you face a new CVE, one of the things you need is knowing which of your supported version are affected.
So you need to look at code history, git blame and so on to determine at what time the vulnerability was introduced.
Having this date tells you what supported version are vulnerable, but also what unsupported version are, without any extra work.
That being determined, moving an old version to another repo or directory or flagging it as vulnerable should be no big deal.

old-releases.md
s0undt3ch: The bootstrap script, in its current incarnation is already bloated and hard to maintain and worse to test.
On the works is a PoC about a python project capable of bootstrapping Salt(single binary using pyinstaller) which doesn't require any system dependencies to bootstrap.
On the new incarnation, we can add such a warning, on the current incarnation, I'm hesitant since the next stable release after Sodium(3001), Magnesium(3002) should be using either the matured PoC or an alternative since a lot will change with Salt moving also to a single binary approach.

bryceml:
>    1. Consider having both retired, and current releases under the new domain.  CI systems, may need to hardcode/template the URL to a specific version. Currently, that is possible with for example: `https://repo.saltstack.com/py3/redhat/$releasever/$basearch/archive/<VERSION>`. The repo RPMs won't work, but I think that is fine.
>        Put another way, I am suggesting to keep the current structure (or similar) under the new domain, and then only have current releases under `repo.saltstack.com`.
> 
>    2. Additionally, rework the following menu, by indicating current releases and no longer supported ones. For example, 2019 is previous, but is still supported.
> 
> 
> image
> 
>    3. On a UX note, consider turning the version drop down menu into a tabs. Similar to how https://packages.ubuntu.com/ does it. All current releases are listed at the top at all times. That saves users a extra click and speeds up interaction, while indicating the status of a release at a glance. Could be color coded, or somehow separated.

@rossengeorgiev 

We believe doing item 1 would cause more harm than good.  The main point of this SEP is to make it harder to use insecure/unsupported versions, I added a comment https://github.com/saltstack/salt-enhancement-proposals/pull/28/commits/dd7df98ad7b8b69afc9845027b9c185ae0ba5bb3 to the sep that addresses it.  Another alternative when this is necessary is to host your own copy of the repo by mirroring the parts you need: https://repo.saltstack.com/#mirror  s3 sync would still be available on archive.repo.saltstack.com.

on items 2 and 3, a redesign is in the works from what I've been told.  That will hopefully address these concerns.
@saltstack/team-docs can verify that.

rossengeorgiev:
hi @bryceml, thanks for you comment. I appreciate the goal of the SEP, and steering toward keeping salt up to date. In reality, for me that is slightly more complicated. I should be able to template my url, and get the version I want, even if its unsupported or insecure. This simplifies a number of workflows. I don't think that in anyway conflicts with the goal of this SEP. Simply, remove the minor version link/option from the `repo.saltstack.com` landing page, and don't advertise that as an option. Then just link to the archive as way to access older versions.

waynew:
This SEP has received the required approvals, and is now an accepted SEP.

hajdukd:
Packages in Archive repository are broken, still point to old repository.

bryceml:
> Packages in Archive repository are broken, still point to old repository.

Which URLs/repos?

hajdukd:
I've just read that the salt team is not willing to repackage repo rpms. And those were the ones I was talking about.

PR #30, opened 2020-06-15
SEP 22:  Compartmentalize salt: Use pop-grains (grainsv2) for grain collection

Akm0d:
Grains is currently a static component of salt.  We want it to be developed in a more compartmentalize fashion.
The engineers at SaltStack have created platform-specific projects that functionally replace the grains provided by salt with POP.  We propose that this project, codenamed `grainsv2` becomes available in salt while maintaining backwards compatibility.

### DIFF
--- a/0022-compartmentalize-grains-via-pop.md
+++ b/0022-compartmentalize-grains-via-pop.md
@@ -33,17 +33,17 @@ individual idem-platform projects independently from salt.
 Grains are currently created by calling functions in the salt loader.  We would simply need to call
 external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
 
-- All salt grains have already been ported to and fully tested on the `grainsv2`.
+- All salt grains have already been ported to `grainsv2`.
 - `grainsv2` has many more features than the salt-grains implementation, but is compatible with salt grains.
 - None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
 - This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
 platform -- making salt easier to develop
-- Grains in the new platform are independently and fully tested.
+- Grains in the new platform are independently fully tested.
 - Development will freeze on core.py and the salt/grains folder.
 - This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
-in early so that any unintentional issues can be resolved quickly.
+in early so that any unintentional issues can be reet up a conversation with the relevant parties solved quickly.
 - Salt releases will aggressively pin `grainsv2`/idem-platform dependencies with each release.
-- Users will be to update their grains implementation to a custom idem-grain version.
+- Users will be to update their grains implementation to a custom idem-platform version.
 - The use of `grainsv2` will be gated with a config variable; It will be disabled by default.
 - New development for grains will be done in `grainsv2` projects and core.py will remain static.
 - When `grainsv2` gains maturity in salt on the other side of the gate (and python3.5 is dropped) it will completely replace core.py
--- a/0022-compartmentalize-grains-via-pop.md
+++ b/0022-compartmentalize-grains-via-pop.md
@@ -39,7 +39,7 @@ external POP functions to collect grains.  We need to maintain backwards compati
 - This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
 platform -- making salt easier to develop
 - Grains in the new platform are independently and fully tested.
-- Development will freeze on core.py it's tests could be skipped for speed improvements.
+- Development will freeze on core.py and the salt/grains folder.
 - This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
 in early so that any unintentional issues can be resolved quickly.
 - Salt releases will aggressively pin `grainsv2`/idem-platform dependencies with each release.
--- a/0022-compartmentalize-grains-via-pop.md
+++ b/0022-compartmentalize-grains-via-pop.md
@@ -1,15 +1,15 @@
 - Feature Name: compartmentalize-grains-via pop
 - Start Date: 2020-06-22
 - SEP Status: Draft
-- SEP PR: (leave this empty)
-- Salt Issue: (leave this empty)
+- SEP PR: https://github.com/saltstack/salt/pull/57681
+- Salt Issue: https://github.com/saltstack/salt/issues/57680
 
 # Summary
 [summary]: #summary
 
 Grains is currently a static component of salt.  We want it to be developed in a more compartmentalize fashion.
-The engineers at saltstack have created kernel-specific projects that functionally replace the grains
-provided by salt with POP.  We propose that this project, codenamed `corn` replaces grains in salt while
+The engineers at SaltStack have created platform-specific projects that functionally replace the grains
+provided by salt with POP.  We propose that this project, codenamed `grainsv2` becomes available in salt while
 maintaining backwards compatibility.
 
 # Motivation
@@ -25,7 +25,7 @@ This also solves the issue of `core.py` being an unmaintainable monolith.  It al
 to be independently maintained.
 
 Working group leaders and other community members focused on specific platforms can manage the release cycle for
-individual grains projects independently from salt.
+individual idem-platform projects independently from salt.
 
 # Design
 [design]: #detailed-design
@@ -33,17 +33,20 @@ individual grains projects independently from salt.
 Grains are currently created by calling functions in the salt loader.  We would simply need to call
 external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
 
-- All salt grains have already been ported to and tested on the new platform.
-- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
+- All salt grains have already been ported to and fully tested on the `grainsv2`.
+- `grainsv2` has many more features than the salt-grains implementation, but is compatible with salt grains.
 - None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
 - This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
 platform -- making salt easier to develop
 - Grains in the new platform are independently and fully tested.
-- Most grains tests will be able to be removed from salt; speeding up it's test suite.
+- Development will freeze on core.py it's tests could be skipped for speed improvements.
 - This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
 in early so that any unintentional issues can be resolved quickly.
-- Salt releases will be able to pin idem-grain dependencies with each release.
+- Salt releases will aggressively pin `grainsv2`/idem-platform dependencies with each release.
 - Users will be to update their grains implementation to a custom idem-grain version.
+- The use of `grainsv2` will be gated with a config variable; It will be disabled by default.
+- New development for grains will be done in `grainsv2` projects and core.py will remain static.
+- When `grainsv2` gains maturity in salt on the other side of the gate (and python3.5 is dropped) it will completely replace core.py
 
 ## Alternatives
 [alternatives]: #alternatives
--- a/compartmentalize-grains-via-pop.md
+++ b/compartmentalize-grains-via-pop.md
@@ -1,0 +1,71 @@
+- Feature Name: compartmentalize-grains-via pop
+- Start Date: 2020-06-22
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Grains is currently a static component of salt.  We want it to be developed in a more compartmentalize fashion.
+The engineers at saltstack have created kernel-specific projects that functionally replace the grains
+provided by salt with POP.  We propose that this project, codenamed `corn` replaces grains in salt while
+maintaining backwards compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+"The original implementation of grains was not designed to scale to the level that salt requires" - thatch45
+Tom wrote POP to solve this problem and we're going to use it. 
+
+By using Plugin Oriented Programming, grains can be separated into individual projects that can be maintained
+by the community safely.
+
+This also solves the issue of `core.py` being an unmaintainable monolith.  It allows platform specific grains
+to be independently maintained.
+
+Working group leaders and other community members focused on specific platforms can manage the release cycle for
+individual grains projects independently from salt.
+
+# Design
+[design]: #detailed-design
+
+Grains are currently created by calling functions in the salt loader.  We would simply need to call
+external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
+
+- All salt grains have already been ported to and tested on the new platform.
+- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
+- Grains in the new platform are independently and fully tested.
+- Most grains tests will be able to be removed from salt; speeding up it's test suite.
+- This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
+in early so that any unintentional issues can be resolved quickly.
+- Salt releases will be able to pin idem-grain dependencies with each release.
+- Users will be to update their grains implementation to a custom idem-grain version.
+
+## Alternatives
+[alternatives]: #alternatives
+
+Keep things the way they are.  Does this mean we would suffer in stagnation?  Will we be stuck maintaining a monolith.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+What will further adoption of POP mean inside of salt?  The purpose of this sep and these features is not to provide
+a definitive answer to this question -- rather to introduce plugin oriented programming into salt in a very safe way.
+In the future this will determine the design so that salt can be further compartmentalized and written in POP.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Please consider:
+
+- We will be introducing a number of new dependencies.
+- The new implementation of grains/pop is written with python3.6 and later with asyncio
+(The latest release of salt is also python3.6 and later).
+- Some grains for specific OSes (such as SUSE) implement some grains differently from other platforms.  We need
+to set up a conversation with the relevant parties to resolve such conflicts.
+- Documentation will be the heaviest lifting.  We need to document how to make salt-style internal grains and generally
+available idem-grains.


compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+In the future this will determine the design so that salt can be further compartmentalized and written in POP.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Please consider:
+
+- We will be introducing a number of new dependencies.
+- The new implementation of grains/pop is written with python3.6 and later with asyncio
+(The latest release of salt is also python3.6 and later).
```
s0undt3ch, line 67: Which release is this?
We currently support 3.5, will Magnesium be >= 3.6?

Akm0d: It would have to be >=python3.6 to support pop code, this is something we should discuss with the team in greater detail.  @dmurphy18 What is tying us to python 3.5?

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+individual grains projects independently from salt.
+
+# Design
+[design]: #detailed-design
+
+Grains are currently created by calling functions in the salt loader.  We would simply need to call
+external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
+
+- All salt grains have already been ported to and tested on the new platform.
+- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
```
mchugh19, line 37: Will the current methods of setting grains be supported in corn? 

1. `_grains` directory
2. grains in minion config
3. grains in `/etc/salt/grains` & the grain modules which modify grains

Akm0d: Yes, grains will be able to be added to salt the same ways they always have.   
You will also be able to add grains the new pop/idem way (such as pip installing a project like [corn cowsay](https://pypi.org/project/corn-cowsay) into the same python environment as salt.

Akm0d:
Opened a PR to start making the change: https://github.com/saltstack/salt/pull/57681

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
+- Grains in the new platform are independently and fully tested.
+- Most grains tests will be able to be removed from salt; speeding up it's test suite.
+- This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
+in early so that any unintentional issues can be resolved quickly.
+- Salt releases will be able to pin idem-grain dependencies with each release.
+- Users will be to update their grains implementation to a custom idem-grain version.
```
daks, line 46: s/will be/will be able/ maybe?

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+- Start Date: 2020-06-22
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Grains is currently a static component of salt.  We want it to be developed in a more compartmentalized fashion.
+The engineers at saltstack have created kernel-specific projects that functionally replace the grains
```
OrangeDog, line 11: Does this mean OS kernel, or is it another food-based pun?

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+[motivation]: #motivation
+
+"The original implementation of grains was not designed to scale to the level that salt requires" - thatch45
+Tom wrote POP to solve this problem and we're going to use it. 
+
+By using Plugin Oriented Programming, grains can be separated into individual projects that can be maintained
+by the community safely.
+
+This also solves the issue of `core.py` being an unmaintainable monolith.  It allows platform specific grains
+to be independently maintained.
```
OrangeDog, line 25: So could simply not having them all in a single file.

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+- All salt grains have already been ported to and tested on the new platform.
+- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
+- Grains in the new platform are independently and fully tested.
+- Most grains tests will be able to be removed from salt; speeding up it's test suite.
+- This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
+in early so that any unintentional issues can be resolved quickly.
+- Salt releases will be able to pin idem-grain dependencies with each release.
```
OrangeDog, line 45: What does "idem" mean in this context?

OrangeDog:
Is there any code in salt core that depends on specific grains?

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Please consider:
+
+- We will be introducing a number of new dependencies.
+- The new implementation of grains/pop is written with python3.6 and later with asyncio
+(The latest release of salt is also python3.6 and later).
+- Some grains for specific OSes (such as SUSE) implement some grains differently from other platforms.  We need
```
OrangeDog, line 68: This sounds like the biggest problem. Even though you say splitting them up has made them more consistent, it would be a lot easier to maintain if they were all in the same place.

How is this going to work for grains whose implementations differ not based on OS? For example based on whether particular services, package managers, or virtualisations are being used? There's a big matrix of how the same grain should get its data on different minions.

compartmentalize-grains-via-pop.md
dmurphy18: @Akm0d Debian 9, Ubuntu 16.04, SLES are the last supported platforms still using Py 3.5

damon-atkins:
I have looked at the structure of corn.
* idem-aix
* idem-bsd
* idem-solaris
* idem-dawwin
* idem-linux

corn/grain fqdn is the same on all platforms, why write the code 5 times?

Why is there a idem-linux vs idem-rhel and idem-debian etc I assume to save having the same code in multiple places.

Why not a idem-posix This is well documented standard which Linux/Unix is based on, and then anything outside this standard then use idem-solaris, etc.

Why not have idem-py i.e. anything which can be completed in pure python e.g. fqdn.  idem-py would exclude the use of posix functions like chmod(), chown() as they are Unix based  and hence not 100% multi-platform.

OrangeDog:
@damon-atkins "pure python" describes non-core packages that are written only in python. It has nothing to do with whether it uses platform-specific functions.

compartmentalize-grains-via-pop.md
Akm0d: It means OS kernel :+1:

compartmentalize-grains-via-pop.md
Akm0d: It also opens the door to future exec modules and states to be added to idem-windows and other platform-specific projects.  That way they you don't need to worry about tests for all OSes passing when your exec module only affects windows.

compartmentalize-grains-via-pop.md
Akm0d: damon's solution below is perfect for this.  Something like an `idem-posix` that contains all  shared code will keep maintainability simple.

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+Grains is currently a static component of salt.  We want it to be developed in a more compartmentalized fashion.
+The engineers at saltstack have created kernel-specific projects that functionally replace the grains
+provided by salt with POP.  We propose that this project, codenamed `corn` replaces grains in salt while
+maintaining backwards compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+"The original implementation of grains was not designed to scale to the level that salt requires" - thatch45
+Tom wrote POP to solve this problem and we're going to use it.
```
waynew, line 19: I think the biggest question here: what scale are we looking for - how do we know if corn/pop-grains meets this required scale?

compartmentalize-grains-via-pop.md
waynew: :point_up: 

I would be curious to see a comparison between these approaches.

compartmentalize-grains-via-pop.md
OrangeDog: I don't see how? This isn't about duplicating shared code, it's about making sure different code stays consistent.

And then how do you find which project is implementing your grain? Is it in `idem-posix` or overridden by `idem-linux` or by `idem-debian` or `idem-ubuntu` or `idem-ec2` or `idem-networkmanager` etc?

It's far easier to get everything right if everything for a single grain is in a single location.

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+Tom wrote POP to solve this problem and we're going to use it. 
+
+By using Plugin Oriented Programming, grains can be separated into individual projects that can be maintained
+by the community safely.
+
+This also solves the issue of `core.py` being an unmaintainable monolith.  It allows platform specific grains
+to be independently maintained.
+
+Working group leaders and other community members focused on specific platforms can manage the release cycle for
+individual grains projects independently from salt.
```
waynew, line 28: I definitely support this idea. I know libraries like flask support this kind of thing with the flask.ext namespace. I would like to see some type of discussion around this approach - maybe in the alternatives section? What tradeoffs are we making going with pop/corn vs the ext approach?

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+[design]: #detailed-design
+
+Grains are currently created by calling functions in the salt loader.  We would simply need to call
+external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
+
+- All salt grains have already been ported to and tested on the new platform.
+- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
```
waynew, line 40: What is the risk of API drift? That is, can particular versions Salt and Corn become incompatible?

Akm0d: Salt will agressively pin a version of corn and all the implementations of corn.  That said, with this new method, users could have a stable version of salt but a development version of windows grains.

compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,71 @@
...
+[alternatives]: #alternatives
+
+Keep things the way they are.  Does this mean we would suffer in stagnation?  Will we be stuck maintaining a monolith.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+What will further adoption of POP mean inside of salt?  The purpose of this sep and these features is not to provide
+a definitive answer to this question -- rather to introduce plugin oriented programming into salt in a very safe way.
+In the future this will determine the design so that salt can be further compartmentalized and written in POP.
```
waynew, line 58: I think we also need to determine how this will be gated. And what happens if this thing goes completely sideways? How can people roll back to the existing way?

compartmentalize-grains-via-pop.md
waynew: One of the things that we've also been seeing -- at least I have -- is something that we had in the past but definitely haven't been keeping up, and that's maturity.

Do we have any kind of formalized patterns around maturity for corn grains? I think this will become *essential* as we explore pop-style dependencies. Some users will be comfortable with rapidly changing and experimental grains/features, others will definitely only want to use mature/stable features that aren't rapidly changing.

compartmentalize-grains-via-pop.md
Akm0d: I think that gating the use of new grains is an excellent compromise.  It means OSes that still require python3.5 can get the current status of grains, but all new development could go into the gated code until python3.5 is deprecated

compartmentalize-grains-via-pop.md
OrangeDog: That doesn't answer the question. Why is the Latin for "same" being used? I assume it doesn't mean "same-grain" or it would just say that.

Akm0d: `[idem](https://gitlab.com/saltstack/pop/idem)` is short for "idempotent".  It is the evolution of the execution and state module system from salt

compartmentalize-grains-via-pop.md
Akm0d: I'm going to add that to the SEP, we will keep core.py as it is; no changes and it won't be removed for a year.  No PRs will be accepted into it and all development of grains will be shifted into pop-grains.  new grains will be gated by a config option.

compartmentalize-grains-via-pop.md
OrangeDog: This isn't a poppadom/pop-idem pun is it? \*squints\*

Akm0d: idempotent
[ ahy-duh m-poht-nt, id-uh m- ] 
(computing) Describing an action which, when performed multiple times, has no further effect on its subject after the first time it is performed.

compartmentalize-grains-via-pop.md
Akm0d: Grains in grainsv2 aren't easily overwritten.  We thought of this early on and each grain can only be set once.  If `idem-debian` tries to overwrite a grain an error will be thrown.  It's certainly  *possible*, but not easy,  to overwrite a grain.

As for maturity, the new grains will be gated and disabled by default.  Users that enjoy the stability of grains as they know and love them will still be able to have that.  Grainsv2 will be able to gain maturity through those who choose to enable them.

I will also note that grains on platforms like BSD, solaris, and AIX, are fully tested in their idem-platform projects and have absolutely no running tests in salt.

compartmentalize-grains-via-pop.md
Akm0d: @dmurphy18 I handled this in the code.  Grainsv2 won't be included in python3.5 and lower builds.  The `__hub__` will also not be available on <=python3.5

0022-compartmentalize-grains-via-pop.md
```diff
@@ -1,15 +1,15 @@
...
+- SEP PR: https://github.com/saltstack/salt/pull/57681
+- Salt Issue: https://github.com/saltstack/salt/issues/57680
 
 # Summary
 [summary]: #summary
 
-Grains is currently a static component of salt.  We want it to be developed in a more compartmentalized fashion.
-The engineers at saltstack have created kernel-specific projects that functionally replace the grains
-provided by salt with POP.  We propose that this project, codenamed `corn` replaces grains in salt while
+Grains is currently a static component of salt.  We want it to be developed in a more compartmentalize fashion.
```
OrangeDog, line 10: That was right the first time

0022-compartmentalize-grains-via-pop.md
```diff
@@ -25,25 +25,28 @@ This also solves the issue of `core.py` being an unmaintainable monolith.  It al
...
-- All salt grains have already been ported to and tested on the new platform.
-- The new platform has many more features than the salt-grains implementation, but is compatible with salt grains.
+- All salt grains have already been ported to and fully tested on the `grainsv2`.
+- `grainsv2` has many more features than the salt-grains implementation, but is compatible with salt grains.
 - None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
 - This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
 platform -- making salt easier to develop
 - Grains in the new platform are independently and fully tested.
-- Most grains tests will be able to be removed from salt; speeding up its test suite.
+- Development will freeze on core.py it's tests could be skipped for speed improvements.
```
OrangeDog, line 42: its

compartmentalize-grains-via-pop.md
Akm0d: It's also about having grainsv2 be tested thoroughly on each platform, having a much faster cadence/release cycle for grains, having working group/community maintained projects that implement platform-specific grains/states/exec modules, speeding up the test suite, speeding up grains.... Breaking up core.py is just one piece of the puzzle

compartmentalize-grains-via-pop.md
max-arnold: A couple of additional compatibility questions:

1. What about other ways to add custom grains (other than `_grains`): `salt-call --module-dirs DIR`, [`grains_dirs`](https://github.com/saltstack/salt/issues/32951) setting, [setuptools entry points](https://docs.saltstack.com/en/latest/topics/tutorials/packaging_modules.html#external-modules-setuptools-entry-points-support)?

2. Will a custom grain that accepts the optional `grains` argument get the core grains passed in? See this feature that was introduced in Salt Fluorine: https://github.com/saltstack/salt/pull/47372

max-arnold: Also, is there a plan to sync corn grains with the recent changes in Salt core grains? For example, the recent FQDNs grain change and the new setting that disables it on some platforms https://github.com/saltstack/salt/pull/55581

Maybe a core grain code freeze period could help?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+
+Grains is currently a static component of salt.  We want it to be developed in a more compartmentalize fashion.
+The engineers at SaltStack have created platform-specific projects that functionally replace the grains
+provided by salt with POP.  We propose that this project, codenamed `grainsv2` becomes available in salt while
+maintaining backwards compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+"The original implementation of grains was not designed to scale to the level that salt requires" - thatch45
```
dwoz, line 18: My impression is that scaling issues currently revolve around the event bus and job storage, not the grains implementation.  Can we elaborate on what we mean by scale here? How does the current implementation fail to scale? How does the proposed change make things more scalable?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+Tom wrote POP to solve this problem and we're going to use it. 
+
+By using Plugin Oriented Programming, grains can be separated into individual projects that can be maintained
+by the community safely.
+
+This also solves the issue of `core.py` being an unmaintainable monolith.  It allows platform specific grains
```
dwoz, line 24: Would and alternate method be to simply split up core.py into multiple files?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+Working group leaders and other community members focused on specific platforms can manage the release cycle for
+individual idem-platform projects independently from salt.
+
+# Design
+[design]: #detailed-design
+
+Grains are currently created by calling functions in the salt loader.  We would simply need to call
+external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
+
+- All salt grains have already been ported to `grainsv2`.
```
dwoz, line 36: Are we sure that all of the changes that have taken place in recent releases are also present in the current pop version?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+- `grainsv2` has many more features than the salt-grains implementation, but is compatible with salt grains.
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
```
dwoz, line 39: Which bugs specifically?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+platform -- making salt easier to develop
+- Grains in the new platform are independently fully tested.
+- Development will freeze on core.py and the salt/grains folder.
+- This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
+in early so that any unintentional issues can be reet up a conversation with the relevant parties solved quickly.
+- Salt releases will aggressively pin `grainsv2`/idem-platform dependencies with each release.
+- Users will be to update their grains implementation to a custom idem-platform version.
+- The use of `grainsv2` will be gated with a config variable; It will be disabled by default.
```
dwoz, line 47: I think we should discuss how we gate this change. Adding a config value that will be rendered un-needed in a future release doesn't seem like a great approach to me.

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+- New development for grains will be done in `grainsv2` projects and core.py will remain static.
+- When `grainsv2` gains maturity in salt on the other side of the gate (and python3.5 is dropped) it will completely replace core.py
```
dwoz, line 49: What indicators will be used to gauge the `maturity` of this new version? How can we ensure that we're not decreasing stability of salt? What tests need to be run?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Alternatives
+[alternatives]: #alternatives
+
+Keep things the way they are.  Does this mean we would suffer in stagnation?  Will we be stuck maintaining a monolith.
```
dwoz, line 54: I don't think it is fair to say the only alternative is maintaining a monolith. I can think of several other possibilities.

- Simply breaking up core.py (the monolith) into multiple files inside the salt codebase.
- Creating separate projects for grains does not inherently *require* the addition of POP. This could be done with simple projects using python's build in mechanisms allowing for plugin type projects. One example of this is pytest, django also has been doing this for a long time.

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+What will further adoption of POP mean inside of salt?  The purpose of this sep and these features is not to provide
```
dwoz, line 59: While the purpose of this sep does not intend to provide an answer to this question it certainly seems to provide an opinion about it. I feel it would be pertinent to this discussion, especially because the current PR implementing this change is adding a lot more than just a grains specific library. The addition of idem-* means *would be* adding idem and POP into salt core.

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+a definitive answer to this question -- rather to introduce plugin oriented programming into salt in a very safe way.
+In the future this will determine the design so that salt can be further compartmentalized and written in POP.
```
dwoz, line 61: Have we had a discussion with the core team and community at large around how we want to compartmentalize salt or has this decision been made (by a small group) already?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Please consider:
+
+- We will be introducing a number of new dependencies.
+- The new implementation of grains/pop is written with python3.6 and later with asyncio
```
dwoz, line 69: I feel the asyncio portion also warents further exploration and discussion. Especially considering the challenges we've had trying to upgrade tornado to a version which uses asyncio.

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+(The latest release of salt is also python3.6 and later).
+- Some grains for specific OSes (such as SUSE) implement some grains differently from other platforms.  We need
```
dwoz, line 71: This seems like we are adding a fair amount management and community synchronization overhead. Are we sure we have thought through all of this and developed the processes needed to deal with it?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+to set up a conversation with the relevant parties to resolve such conflicts.
+- Documentation will be the heaviest lifting.  We need to document how to make salt-style internal grains and generally
```
dwoz, line 73: Specifically what documentation are we lacking now. What is the plan and time-lines around getting all of the needed documentation in place?

dwoz:
> I have looked at the structure of corn.
> 
> * idem-aix
> * idem-bsd
> * idem-solaris
> * idem-dawwin
> * idem-linux
> 
> corn/grain fqdn is the same on all platforms, why write the code 5 times?
> 
> Why is there a idem-linux vs idem-rhel and idem-debian etc I assume to save having the same code in multiple places.
> 
> Why not a idem-posix This is well documented standard which Linux/Unix is based on, and then anything outside this standard then use idem-solaris, etc.
> 
> Why not have idem-py i.e. anything which can be completed in pure python e.g. fqdn. idem-py Would exclude the use of posix functions like chmod(), chown() as they are Unix based and hence not 100% multi-platform like most of python.

Is this comment going to be addressed?

compartmentalize-grains-via-pop.md
dwoz: I'm definitely not a big fan of having another dunder added to salt. It seems like sprawl in lack of a properly architected design change.

compartmentalize-grains-via-pop.md
Akm0d: Yes, a core grain code freeze is part of this SEP.  We don't want to maintain two forks of grains and all future development of grains should go into the compartmentalized projects..

Other ways of adding grains will remain available as they are.  

A custom grain that accepts the optional `grains` argument would get the grains generated by grainsv2.

Yes, the test in the [Grainsv2 PR](https://github.com/saltstack/salt/pull/57681) shows a line by line diff of grainsv2 vs salt grains

0022-compartmentalize-grains-via-pop.md
Akm0d: @thatch45  ^^

0022-compartmentalize-grains-via-pop.md
Akm0d: Grains is the first focus of this SEP and the associated changes.  It also opens the door for getting execution modules and state modules from idem in the future.  For example, Mac specific execution modules could be put in idem-darwin, tested in idem-darwin, and PRs/issues can be handled by the Mac working group.  Grainsv2 is the simplest and first step towards more compartmentalization of salt in this way.

0022-compartmentalize-grains-via-pop.md
Akm0d: I have been focused on idem-cloud since March, there will need to be a code freeze of future PRs into core.py and changes in the past two months into core.py will need to be merged into the new platform.

0022-compartmentalize-grains-via-pop.md
dwoz: This is one of the thing I think that should be highlighted more in the SEP. We're saying this is opening the door to door to all of these changes, but we're also saying 
```
What will further adoption of POP mean inside of salt?  The purpose of this sep and these features is not to provide a definitive answer to this question
```
Which is it?

0022-compartmentalize-grains-via-pop.md
Akm0d: No specific open issues, but a lot of work has been done to standardize grains between the different idem platforms.  For example, idem-windows implements the `oscodename` grain and the `gpus` grain, which are missing on windows in salt.   grainsv2 is more complete and consistent across all platforms

0022-compartmentalize-grains-via-pop.md
Akm0d: You're right, this SEP and the accompanying PR do provide strong opinions on this, I mean to say that the future of idem in POP in salt has been discussed by a small group, but warrants a larger discussion, more firm plans, and possibly it's own SEP.   With this [PR](https://github.com/saltstack/salt/pull/57681) we add the `__hub__` to the loader, which means custom execution modules and states could make use of idem's hub... but we have discussed adding the option to specify the salt/idem engine in salt states (defaulting to the salt engine)  is a possible future integration of idem into salt.  This SEP opens the door for that discussion, but that is an implementation/discussion that should happen separately from grains.

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+
+Grains are currently created by calling functions in the salt loader.  We would simply need to call
+external POP functions to collect grains.  We need to maintain backwards compatibility with custom user grains.
+
+- All salt grains have already been ported to `grainsv2`.
+- `grainsv2` has many more features than the salt-grains implementation, but is compatible with salt grains.
+- None of the grain names have been changed, but the output of grains is more consistent/complete between platforms.
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
+- Grains in the new platform are independently fully tested.
```
dwoz, line 41: Are they tested against current versions of salt?

0022-compartmentalize-grains-via-pop.md
Akm0d: Compartmentalizing salt is a discussion that has been had in small groups, and it will happen in larger groups before decisions are made

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+provided by salt with POP.  We propose that this project, codenamed `grainsv2` becomes available in salt while
+maintaining backwards compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+"The original implementation of grains was not designed to scale to the level that salt requires" - thatch45
+Tom wrote POP to solve this problem and we're going to use it. 
+
+By using Plugin Oriented Programming, grains can be separated into individual projects that can be maintained
```
dwoz, line 21: Are there other alternative approaches we could take instead of 'Plugin Oriented Programming'?

0022-compartmentalize-grains-via-pop.md
Akm0d: I think this is an excellent point and exposes the crux of what needs to be discussed in this SEP.   Broadly I've speculated that the windows working group would manage idem-windows and the mac working group would manage idem-darwin and at a certain point in the release there would be a discussion about which version of idem-darwin and idem-windows would be pinned to the next release of salt.  But we do need a concrete process in place.

0022-compartmentalize-grains-via-pop.md
dwoz: Seems like the addition of `__hub__` is inviting a lot of confusion and potential complexity to and already complex code base in regards to the loader. What problem is the addition of `__hub__` intended to solve? How will developers know which one to use? What happens if we mix the use of `__salt__` and `__hub__`?

0022-compartmentalize-grains-via-pop.md
Akm0d: For grains there isn't much more documentation changes that will need to happen, but for the future of the compartmentalization of salt this is an unaddressed problem.    Execution and state modules could be maintained in platform specific projects...  How do we merge that documentation into salt?

0022-compartmentalize-grains-via-pop.md
Akm0d: yes, this [PR](https://github.com/saltstack/salt/pull/57681) has a test that runs both versions of grains and does a line by line comparison of each

0022-compartmentalize-grains-via-pop.md
taigrr: For something as important as grains to be changing, if like to take a more in depth look at POP in salt to understand the significance before a major migration is undertaken.

0022-compartmentalize-grains-via-pop.md
taigrr: For something as important as grains to be changing, like to take a more in depth look at POP in salt to understand the significance before a major migration is undertaken.

compartmentalize-grains-via-pop.md
dwoz: I think we should at least explore if there are simpler, less complex solutions.

0022-compartmentalize-grains-via-pop.md
OrangeDog: Have PRs been opened to fix these things in the current grains?

0022-compartmentalize-grains-via-pop.md
taigrr: Farming out implementations to outside of salt itself h bring up new continuity issues when there's no central authority to determine if for example Ubuntu or Void has the "official/accepted/blessed" format/implementation.

(Examples picked for systemd disagreement reasons)

0022-compartmentalize-grains-via-pop.md
Akm0d: > > I have looked at the structure of corn.
> > 
> > * idem-aix
> > * idem-bsd
> > * idem-solaris
> > * idem-dawwin
> > * idem-linux
> > 
> > corn/grain fqdn is the same on all platforms, why write the code 5 times?
> > Why is there a idem-linux vs idem-rhel and idem-debian etc I assume to save having the same code in multiple places.
> > Why not a idem-posix This is well documented standard which Linux/Unix is based on, and then anything outside this standard then use idem-solaris, etc.
> > Why not have idem-py i.e. anything which can be completed in pure python e.g. fqdn. idem-py Would exclude the use of posix functions like chmod(), chown() as they are Unix based and hence not 100% multi-platform like most of python.
> 
> Is this comment going to be addressed?

Pure python grains and idem-posix are great ideas 👍.  We will create an `idem-universal` for pure python grains/exec/state modules and `idem-posix` that has the shared code between posix platforms.  I am especially a fan of not having to push updates to idem's implementation of cmd.run to 7 different platform specific repos.  Code that is similar between platforms should be in universal repos.

0022-compartmentalize-grains-via-pop.md
dwoz: Good point, how would we ensure that a `requires_reboot` grain on one system isn't named `requires_restart` on another system?

0022-compartmentalize-grains-via-pop.md
Akm0d: no

terminalmage: > This will resolve many bugs that are open against grains

conflicts directly with

> No specific open issues

Are there open bugs that this solves, or not? If not, perhaps this SEP should not claim that there are. Regardless, I think this point needs to be clarified.

0022-compartmentalize-grains-via-pop.md
terminalmage: Perhaps a new value isn't necessary. We have `use_superseded`, which was only being used for the format change in `module.run`. Couldn't it be used here?

0022-compartmentalize-grains-via-pop.md
```diff
@@ -0,0 +1,74 @@
...
+- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt
+platform -- making salt easier to develop
+- Grains in the new platform are independently fully tested.
+- Development will freeze on core.py and the salt/grains folder.
+- This comprises a complete rewrite of ALL grains.  Backwards compatibility is expected, but we need this to get
+in early so that any unintentional issues can be reet up a conversation with the relevant parties solved quickly.
+- Salt releases will aggressively pin `grainsv2`/idem-platform dependencies with each release.
+- Users will be to update their grains implementation to a custom idem-platform version.
+- The use of `grainsv2` will be gated with a config variable; It will be disabled by default.
+- New development for grains will be done in `grainsv2` projects and core.py will remain static.
```
terminalmage, line 48: So does this mean that no new core grains will be able to be added to Python 3.5 platforms (Ubuntu 16.04 LTS, etc.)?

0022-compartmentalize-grains-via-pop.md
terminalmage: To build on @dwoz's question, assuming that the platform-specific projects are going to be providing certain grains that used to be in core.py, what are the plans for enforcing that a given platform's `idem-XXXXX` project provides the grains it is supposed to?

Akm0d:
1. We will rescind this SEP for now, it clearly does too many things
2. We want to redo it in a way that allows grains to be generated by an external source, that way we can have a pluggable interface for generating grains (that doesn't mess with the loader) -- and we need to discuss this more
3. For now, the hub will not be part of salt

dwoz:
Note: I already typed this out be the SEP was closed. So, just adding the comment for posterity.

I agree that at least some 'compartmentalization' of salt needs to take place. However, I think the main things we should be concerned with are rarely used, un-maintained, and potentially un-needed execution modules and states. I'm not sold on the idea that we need to compartmentalize core components of the salt code base. The idea around moving execution modules to their own repositories is that the people focused on core can *focus* on the core and not spend time maintaining modules may not be used by a majority of users. Separating out core functionality into smaller code bases seems like it could have the un-desired side-effect of having to split our (anyone interested in contributing to the core's) focus from one place into many places. It will may make the core of salt a lot harder to reason about, and troubleshoot. It might mean that one group of people focused on a specific piece of the core (or platform) ends up making decisions that effect the everyone else without everyone being aware. I feel like we already struggle with this today. 

There are upsides and downsides of monoliths. Anyone who went through the micro-services revolution knows there are also upsides and downsides to having components split up into potentially overly small units.

cachedout:
I spent a few minutes looking at this before I noticed that it had been closed but I think that @dwoz has this exactly right. 

It's very tempting to believe that a repo split would be beneficial but I am deeply skeptical. Specifically, here are some points that I struggle with:

```
Working group leaders and other community members focused on specific platforms can manage the release cycle for
individual idem-platform projects independently from salt.
```
TBH, this sounds like an absolute nightmare for users to maintain and troubleshoot.

Suddenly, Salt users will need to reason not just about a single version of Salt but (n) number of subsystems as well, all of which are on different release cycles? It's hard to imagine a coherent testing strategy for a matrix of compatibility like this and I think it's unlikely that development concerns will remain as siloed as this model assumes. 

While I believe I understand the arguments in favor of this type of compartmentalization, I strongly suspect the practical reality is going to be a bit of mess. I suspect strongly that this is going to be decision that the Salt project regrets. (And it's worth noting that a competing project tried this exact approach in 2016 and [reverted to a mono-repo after a user revolt](https://docs.ansible.com/ansible/2.4/dev_guide/repomerge.html).)

```
- This will resolve many bugs that are open against grains and will migrate those bugs OUT of the salt repo.
```
Again, why is this good? A user now has to reason about the right repo, go there to file a bug, and hope a maintainer is active there? I struggle to imagine a world in which this is better experience for a user. As I have suggested repeatedly, a better solution for this problem would be to facilitate the further participation of working groups and the Salt community inside the mono-repo.

plinkable:
Having missed all the fun and only read all this after it was closed, I have a simple suggestion instead: create a way to build component packages out of the single repo.

Reviewing and maintaining changes would be in a single place, and individual component release cycles could be managed easily.

Major salt releases would then include/reference already-published-and-tested components, rather than building the bleeding edge that hasn't felt like a stable release for some time now.

PR #32, opened 2020-08-12
SEP 24: Add sub_state_return

Akm0d:
Add the ability for states to have `sub_state_run`s so that state output from modules that can execute multiple states/SLSs/configurations/plays simultaneously can very clear and verbose.

### DIFF
--- a/0024-sub-state-returns.md
+++ b/0024-sub-state-returns.md
@@ -38,8 +38,6 @@ Definitions
 - External state engine: A salt module that has the ability to execute a group of configurations (I.E Ansible, Chef, Idem)
 - Sub state run: A new key in the return of salt states that allows definitions of external state engine run details
 
-This is the bulk of the SEP. Explain the design in enough detail for somebody familiar
-with the product to understand, and for somebody familiar with the internals to implement. It should include:
 
 How it is used
 --------------
--- a/0024-sub-state-returns.md
+++ b/0024-sub-state-returns.md
@@ -1,4 +1,4 @@
-- Feature Name: Sub State Returns
+- Feature Name: Sub State Runs
 - Start Date: 2020-08-12
 - SEP Status: Draft
 - SEP PR: github.com/saltstack/salt/pull/57993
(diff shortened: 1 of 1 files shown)


OrangeDog:
Should this also cover "internal" sub-states? There are a few states that call `file.managed` internally, for example.

Oloremo:
> Should this also cover "internal" sub-states? There are a few states that call file.managed internally, for example.

Yeah, we wrote an abomination of the state that calls for 4 different states and we have to play around this very hard to be able to return correct results.

Akm0d:
The original motivation for this was indeed because of idem and idem-cloud.
The first connections I drew were to Ansiblegate and chef, but for sure, states that call `file.managed` or other states internally would be able to show the changes made by those individual state calls.  So yes, this should cover "internal sub states" thank you for that insight :+1:

0024-sub-state-returns.md
```diff
@@ -0,0 +1,153 @@
...
+            "start_time": None,  # Optional, the timestamp of the external state run's start time
+            "low": {
+                "name": None,        # The name of the state from the external state run
+                "state": None,       # Name of the external state run
+                "__id__": None,      # ID of the external state run
+                "fun": None,         # The Function name from the external state run
+            },
+        }
+    ],
+}
```
Akm0d, line 77: Is this a good way for the return data to be formatted?  I wanted to pass it through the state "low_data" function the same as regular states.  Any thoughts on the implementation here?

waynew: > This talks a lot about the motivation and impact on `idem`, but what about the regular salt engine?
> What are the internal and observable changes to e.g. the ansible states, and would they require e.g. changes to peoples' returners?

By default, they won't see any change. It won't break, but nobody will see anything fancy - it will look exactly how it does today. *But* if the ansible states/modules are updated to actually return the sub state return data? 🎉  then they would show up. 

Basically this proposal is driven by `idem`, but can benefit anything that takes advantage of returning more structured data.

@Akm0d maybe it would also help to add a sample of the current output, as well as what the output would look like with this change applied - obviously using the default outputter, but perhaps JSON style as well.

OrangeDog:
> if the ansible states/modules are updated to actually return the sub state return data

Shouldn't part of the implementation plan be to do that then (and everywhere else applicable)? A PoC and an analysis of e.g. whether it's a breaking change.

Akm0d:
The PoC has been done with [idem](https://github.com/saltstack/salt/pull/57969).  It works and does not appear to be a breaking change. The door would be wide open for sub state returns to be applied everywhere else they would be useful.

OrangeDog:
> The PoC has been done with idem. It works and does not appear to be a breaking change.

Yes, but I'm asking about the main salt engine, not idem.

waynew:
> > if the ansible states/modules are updated to actually return the sub state return data
> 
> Shouldn't part of the implementation plan be to do that then (and everywhere else applicable)? A PoC and an analysis of e.g. whether it's a breaking change.

It's probably worthwhile doing a PoC with Ansible/etc. to confirm whether or not they can take advantage of these changes, and whether or not doing that would be a breaking change, though by adding return data I'm not sure how it would be -- unless one is depending on the *absence* of keys in the return, which seems like a bad idea 😬 

But I don't agree that this SEP/feature should **require** that anything uses it, it should just make the functionality available -- and as you mentioned it shouldn't require breaking changes (unless absolutely necessary).

waynew:
@Akm0d btw apparently I didn't let you know - this is SEP 24, feel free to update your file!

garethgreenaway (approved):

waynew (approved):

PR #34, opened 2020-09-03
SEP 26: Package Salt with Tiamat

bryceml:


krionbsd (approved):

raddessi:
Say I need to patch an issue in salt, is this still possible via some other method after this is packaged in to a frozen binary? I would love to say don't do this but I've had multiple patches for every version of salt that we've used in prod so far.

In the current system I just use salt's own `file.patch` to self-patch.. would we need to say keep our own fork of the branch we're running then apply our patches and repackage it up there, then distribute internally? Or is there another method?

Akm0d:
> In the current system I just use salt's own `file.patch` to self-patch.. would we need to say keep our own fork of the branch we're running then apply our patches and repackage it up there, then distribute internally? Or is there another method?

That's exactly what you would need to do; fork -> apply patch -> repackage

I'm currently building a docker container for tiamat that uses a very old glic for maximum portability (and a second container that builds with wine for easy windows .exe creation)

waynew:
@raddessi the other option is to deploy this single binary as a "onedir" build - salt+all deps get extracted to one directory, so you would be able to patch *those* files.

bryceml:
>  deploy this single binary as a "onedir" build - salt+all deps get extracted to one directory, so yo

I thought the tiamat onedir only had pyc files, no py files.  We should get some clarification there.

Akm0d:
oh onedir builds can have py files, it's the single binary that you can't change

raddessi:
Hey nice to hear! I do like options. That's right I remember Tom talking about the "onedir" builds somewhere before.. maybe saltair? OK, both of those are viable (assuming `.py` files are included). :+1: Thanks all for the details

barneysowood:
The SEP suggests that the single binary will have all required Python Libraries and any C Library deps bundled. Looking at the pyinstaller docs, I'm not totally sure this is correct - I don't think it goes as far as bundling glibc for example does it? Might be good to clarify.

I also think the proposal needs to cover what the process will be for tracking what libraries at what version are bundled and how Saltstack will handle the issue of a security vulnerability in a bundled library.

Is there a way for a tiamat binary to show what libraries (and their versions) are bundled? 

@Akm0d - the work you're doing with trying to build stuff "with a very old glibc" sounds somewhat concerning. What distro are you using and how old is it? I assume it's something that's still getting security updates - otherwise, how are you ensuring you don't end up bundling something with a known vulnerability?

Fundamentally, I think if you're proposing to move to a model where you are asking people to install software with a number of libraries bundled in, in a way that is opaque and not patchable in using normal OS procedures, then it's got to be clear that there's strong processes in place around security.

If you go do this road you're going from being responsible for providing updates for vulnerabilities in Salt to being  responsible for providing updates for vulnerabilities in Salt *and* all it's dependencies.

s0undt3ch (approved):
> - It would be a learning curve for people to understand how to add python libraries to the new environment.

We could expand here that we will likely be providing a `pip` command to install additional packages, so, the process will be almost the same as `pip install package`

> - Adding python libraries to the new environment might be harder in an air gapped network.

A previous `pip download package` on a similar system(because of the arch) using the same python version as the tiamat salt package would facilitate this scenario.
Then the downloaded packages could be brought into the air gapped environment for install. Since the tiamat salt package will include a `pip` command, the `pip download package` could potentially be done by a salt tiamat package with the same version as that in the air gapped system in order not to worry about arch/python version compatibility.

>  - People would have to re-install their python packages potentially if we updated the python in the tiamat environment on upgrades.

We could potentially have a directory which could keep the downloaded wheels/sdist since we will control how things get installed and use that on upgrades; If we can have the tiamat packaged salt detect that it's being upgraded, or by a special command that would reinstall everything on demand.

s0undt3ch:
> I also think the proposal needs to cover what the process will be for tracking what libraries at what version are bundled and how Saltstack will handle the issue of a security vulnerability in a bundled library.

For the salt tiamat package, all shipped dependencies will be locked to a version(and the dependencies of a dependcy too).
Look under `requirements/static/py3.*/linux.txt` on the salt repo for an example(this is what we use for reproducible testing)

> Is there a way for a tiamat binary to show what libraries (and their versions) are bundled?

Since the salt tiamat package will likely have a `pip` command, `pip list` would likely be available.

> If you go do this road you're going from being responsible for providing updates for vulnerabilities in Salt to being responsible for providing updates for vulnerabilities in Salt and all it's dependencies.

That's correct, and a `.x` release would be made whenever needed to address a dependency security issue.
We are screening our dependencies for security advisory issues and we would release updates when needed, or, information on how to proceed to upgrade a dependency.

max-arnold:
A couple of questions:

1. Will minor Salt upgrades require reinstallation of 3rd-party python libraries? Any other potential upgrade issues?
2. How this change will affect Salt branching model and/or release frequency? What is the general policy regarding dependency versions and upgrades? There are plenty of security and non-security package upgrades (distros often supply a couple of patches on top of upstream source code). Any plans to track those for bundled libraries? Should we expect more frequent minor releases due to dependency upgrades? Can the existing tag-based release procedure (as opposed to branch-based) handle the increased minor release frequency?
3. `It also means that we can have a single binary as an option`. Could someone elaborate on this? I assumed that all Tiamat-based packages will be single-binary. Is that not the case? Which platforms will get the bundle and which one the single binary?
4. Is there any performance impact? I suspect that single binary salt-call startup time can suffer. Any other potential issues?
5. `It would be a learning curve for people to understand how to add python libraries to the new environment`. How the new way of adding libraries will look like?

max-arnold:
> > In the current system I just use salt's own `file.patch` to self-patch.. would we need to say keep our own fork of the branch we're running then apply our patches and repackage it up there, then distribute internally? Or is there another method?
> 
> That's exactly what you would need to do; fork -> apply patch -> repackage

@Akm0d I also have this question. A quite common way to work around Salt bugs (and lack of frequent minor bugfix releases) is to self-patch it using the `file.patch` state and a diff file that is distributed through `salt://` fileserver. This is much easier than repackaging (no need to have packaging pipelines for different OSes, manage repositories, etc.). Just install Salt from upstream repo and apply a state that patches it. Will this still be possible with Tiamat-based packages?

OrangeDog:
Just on a personal opinion, one of the most annoying things as a linux admin is projects that bundle all their own dependencies, instead of working properly with the OS package system. I guess the tide is turning against this (for desktop users anyway) with e.g. snaps.

I am very sceptical that the salt team is capable of maintaining security patches for all targets in all the dependencies. Most distros are patching something at least every week, with thousands of people supporting it. Meanwhile we constantly hear from salt how it's not possible to do even minor releases more frequently than every couple of months. This proposal increases the maintenance burden by a huge amount.

I need to see a believable commitment to do this properly, and not just vague promises. How (for example) are you going to get patches for undisclosed major CVEs, the way the distributions do?

bryceml:
@OrangeDog this makes it easier on us to package more often because to install newer dependencies, we just upgrade via pip as part of a pipeline.  The old way was really hard to update any dependencies that we added to the repo (take a look at amazon linux's dependencies in the salt repo) because we have to go find updated source rpms or source debs for each of those and then create the package.  With all those rpm's and deb's, which we often had to create ourselves, it added a lot of manual effort.

Windows and mac we already bundle everything and we can usually create those packages and get them tested in a few hours.  The linux ones take several days.

Hopefully this helps ease your concern.

OrangeDog:
This issue affects regular users, formulas, and custom modules.

Many modules require an additional python dependency in order to work.
You can get it with either `pkg.installed` or `pip.installed` (which may also require `pkg.installed` for dev dependencies).

How is this going to work now?
Are all the formulas going to need to detect whether this is a tiamat installation in order to do the right thing?
Is the default environment for `pip.installed` going to be the system python, or the tiamat one (either way requires people to change something)?
Will system python packages be available in the tiamat environment?
What about compatibility between non-python system packages and those bundled by tiamat?

waynew:
@max-arnold - I think https://github.com/saltstack/salt-enhancement-proposals/pull/34#issuecomment-686769847 answers at least a few of your points.

With a single binary style deployment it's static (though custom _modules & the like would still work -- I can't see why they wouldn't), so patching a state would not be possible (without overriding it with a custom _state, say).

However, with the onedir style deployment (IIUC it would go under `/opt`, though there might be platform specific exceptions) you *can* manually patch files to your heart's content.

IIUC it's possible to take the same build and either execute it or have it unpackage itself into a onedir style deployment, but I might be misunderstanding that.

> A couple of questions:
> 
> 1. Will minor Salt upgrades require reinstallation of 3rd-party python libraries? Any other potential upgrade issues?

With the Tiamat builds, IIUC, the 3rd party libs would be included - though I'm not sure about optional deps. @Akm0d if this isn't documented in the SEP or documented elsewhere and linked from here, it should be.

For source-only deps there shouldn't be an issue, but there may be issues with compiled deps. For sure there will be if the ABI has any changes, but there might not be otherwise.

> 2. How this change will affect Salt branching model and/or release frequency? What is the general policy regarding dependency versions and upgrades?

It should reduce the burden of releasing Salt, since vagaries in dependencies would be eliminated - we'll just ship the deps that Salt needs.

Obviously that means that we would have to ensure that we're keeping up with security advisories and cutting new releases when dependencies have issues. That should definitely be addressed in this SEP.

>  There are plenty of security and non-security package upgrades (distros often supply a couple of patches on top of upstream source code). Any plans to track those for bundled libraries? Should we expect more frequent minor releases due to dependency upgrades?

I believe that [Pedro's answer covers this](https://github.com/saltstack/salt-enhancement-proposals/pull/34#issuecomment-687292534)

> Can the existing tag-based release procedure (as opposed to branch-based) handle the increased minor release frequency?

We *do* already branch if we need to make a .2 release (or worse). Our existing approach to other releases should not have to change.

> 3. `It also means that we can have a single binary as an option`. Could someone elaborate on this? I assumed that all Tiamat-based packages will be single-binary. Is that not the case? Which platforms will get the bundle and which one the single binary?

(see above)

> 4. Is there any performance impact? I suspect that single binary salt-call startup time can suffer. Any other potential issues?

single binary startup time definitely suffers, for the obvious reason of extracting the binary. I'm not aware of any other known issues. I seem to recall hearing something about improvements in the one-dir approach over existing Salt, but I don't remember the specifics.

> 5. `It would be a learning curve for people to understand how to add python libraries to the new environment`. How the new way of adding libraries will look like?

That should definitely be documented and either added to the SEP or linked from the SEP

> 6. Where in the filesystem Tiamat-based Salt will be located? Something like `/opt`? Is it possible to install multiple major versions simultaneously?

For a single binary option it should definitely be possible. I mean, heck, you can already do that *now* with Salt if you install to a virtual environment, or otherwise munge your PATH.

waynew:
> What about compatibility between non-python system packages and those bundled by tiamat?

I'm not sure I understand the question. https://gitlab.com/saltstack/pop/tiamat/-/blob/master/docs/topics/quickstart.rst might help.

OrangeDog:
@waynew that's the probably the same question as what you said about ABI issues.

barneysowood:
> > I also think the proposal needs to cover what the process will be for tracking what libraries at what version are bundled and how Saltstack will handle the issue of a security vulnerability in a bundled library.
> 
> For the salt tiamat package, all shipped dependencies will be locked to a version(and the dependencies of a dependcy too).
> Look under `requirements/static/py3.*/linux.txt` on the salt repo for an example(this is what we use for reproducible testing)

> > If you go do this road you're going from being responsible for providing updates for vulnerabilities in Salt to being responsible for providing updates for vulnerabilities in Salt and all it's dependencies.
> 
> That's correct, and a `.x` release would be made whenever needed to address a dependency security issue.
> We are screening our dependencies for security advisory issues and we would release updates when needed, or, information on how to proceed to upgrade a dependency.

@s0undt3ch - that sounds great and like you're covering the issues raised. Would it be possible to get that documented in the SEP as part of the detailed design? I think it's a fundamental part of what's being proposed so it would be good to have it covered in the SEP.

barneysowood:
> > 1. Is there any performance impact? I suspect that single binary salt-call startup time can suffer. Any other potential issues?
> 
> single binary startup time definitely suffers, for the obvious reason of extracting the binary. I'm not aware of any other known issues. I seem to recall hearing something about improvements in the one-dir approach over existing Salt, but I don't remember the specifics.

Is it possible to quantify this? If calling `salt` or any of the cli tools is going to require the extraction each time that's going to make commandline use clunky.

waynew:
@barneysowood the extraction is transparent to the user, aside from the extra time - a couple of seconds on a reasonably modern system.

But if that's unacceptable, that's why onedir exists 🙃

barneysowood:
> @barneysowood the extraction is transparent to the user, aside from the extra time - a couple of seconds on a reasonably modern system.
> 
> But if that's unacceptable, that's why onedir exists 🙃

@waynew are there any Tiamat builds of 3001.1 available to test with currently?

waynew:
I think that's a question for @bryceml. I know it's possible to build your own packages with Tiamat, and at least beardedeagle (on Slack) has been making their own internal Tiamat Salt builds. I don't know if there are instructions on the Tiamat page on how to build Salt, but the quickstart guide did look pretty straight forward (linked above in one of my other comments).

bryceml:
Doesn't look like it. It should be as easy as tagging the salt-pkg repo and they should show up on the artifactory staging repos.  We can probably do that without much difficulty next week.

sjorge:
> Say I need to patch a bug in salt, is this still possible via some other method after this is packaged in to a frozen binary? I would love to say I don't do this but I've had multiple patches for every version of salt that we've used in prod so far.
> 
> In the current system I just use salt's own `file.patch` to self-patch.. would we need to say keep our own fork of the branch we're running then apply our patches and repackage it up there, then distribute internally? Or is there another method?

Depending on what your patching, can't we also sue extmods? to load new modules and states files from _states/ and _modules/ ?

cachedout:
>  How (for example) are you going to get patches for undisclosed major CVEs, the way the distributions do?

This question by @OrangeDog is extremely important and so far as I can tell, didn't get a direct response. If I understand this proposal correctly, this has the potential to put Salt badly on the back foot if a bad CVE hits for a dependency. I understand that the desire is to respond quickly, but as @OrangeDog correctly points out, distros often have these prepared in advance while Salt will not. This is a potentially huge deal, IMHO, as it has the potential to expose systems to known flaws without any recourse but to wait for a new release of Salt and deploy it across large fleets of machines.

I'm looking here at what I assume would be a list of deps which would be bundled for an OS like Redhat: https://repo.saltstack.com/py3/redhat/7.8/x86_64/3001/ . Is that a reasonable way to determine which deps would be bundled? (It's not clear from reading this proposal.)

I picked python-requests to look at which is _just one dependency_ of Salt since it's both in the requirements files and in the repo I linked above and I presume would be bundled using this new method, correct? It has has 38 CVEs so far during 2020. In 2019, Redhat did at least 10 (and possibly more) updates to the requests package alone.

I see two paths here:

1) You issue releases of Salt for every CVE across all your deps. This seems as if it could lead to dozens of a releases a year at a minimum. I struggle to see how this would be a win for Salt users and doesn't seem sustainable for the release team either.

2) You independently evaluate each CVE to determine whether or not the security risk posed introduces a security risk to Salt. This may be possible though challenging, but it would be good to learn more about how the team plans to achieve this.

So, my questions are as follows:

1) Has anybody undertaken an audit to estimate how many times a year Salt would need to release new packages under this new model? Perhaps 2019 could be used as a way to make an estimation for future years?

2) What would the policy for upstream CVEs be? Would a new release of Salt be released for every CVE discovered in an dependency or would they be independently evaluated to determine if they pose a risk to Salt itself? If the latter, could we discuss more about how this evaluation would happen and how such an evaluation could be done in a timely manner given that no advance notice by the upstream maintainers would be available to the release team?

3) Which releases of Salt would be patched in cases where a security release of Salt is issued due to an upstream CVE? Would it just be the most recent version or would it go back a number of versions?

Thanks!

boltronics:
**Security**
It seem that many were impacted by CVE-2020-11651 and CVE-2020-11652 just a few months ago. It took too long for people to upgrade - and in this instance it was only Salt that had the problem. I suspect that some of that was because people were installing old versions through pip and the like instead of via OS updates, and it seems reasonable that Tiamat builds may further encourage such behavour.

However another possibility is that some people deliberately didn't upgrade sooner because they didn't spend time to update their custom states, etc. to be compatible with at least 2019.2 before it was EOL, and only bothered to upgrade when Salt itself was the issue. With this new Tiamat packaging proposal, that situation is going to be much more dangerous since any of the dependencies could be vulnerable and impact the user, even if there was no Salt-specific CVE. Would the average person who may not follow Salt closely know they need to upgrade to keep safe?

When new packages come out to address security issues with Salt, how will this be advertised? If you need to make security releases multiple times a month, that would reflect poorly upon the Salt project. On the other hand, if it's not clear that it's a security release, it puts users who were not planning on upgrading at risk.

**Low-end hardware**
It remains to be seen how this will impact low-end hardware, such as a Raspberry Pi, with already low memory usage and I/O from an SD card. There's no memory savings to be had from having a separate version of Python and related libraries. It will be interesting to look at what impact this has on these kinds of very low end environments, that members of the community might use for home routers and similar basic requirements, particularly where the device is already running other Python applications.

**Downstream packaging**
It is possible that this will cause distributions such as Debian to no longer bother to maintain their own packages, as presumably this could become much harder for them to continue to do so, in a way that would meet their policy requirements. I personally know people who only use versions of Salt that are provided and supported by their chosen distribution - sometimes because the distribution has a consistent, strict packaging and support policy it adheres to. It's good for users to have this flexibility, and also good for Salt to get the exposure from having distributions support the project, even if they are typically older releases. Salt would be a harder sell for people with this mindset who would never use upstream if they can help it, but see the usual Salt competitors are packaged, supported by the distribution and just one command away from installation.

**Motivation**
Lastly, I would like to address some of the points mentioned in the _motivation_ section of the document. What are the distros that Salt wants to officially support but cannot currently? Is that something SaltStack is wanting to do, or just a hypothetical? If so, which distributions specifically? I personally would like to see Salt work on GNU Guix SD, but I guarantee nobody there would so much as glance at anything build using Tiamat (are Tiamat builds even reproducible?).

If a Tiamat build fails to work on one distribution due to a change in the environment outside the scope of what Tiamat supports (eg. an issue with a distribution-specific environment variable or a system file location, file format, patched system command, etc.), would everyone have to upgrade? Or would there still be distribution-specific repositories? There's nothing worse than seeing a 100Mb+ update and needing to restart the process because a critical bug was fixed for a specific platform that you're not even running.

Regarding the issue that was discussed in the Salt Community Open Hour (where a weekend was wasted because of issues at the last minute packaging a dependency), there are other ways to prevent that scenario. For example, at my company, our software automatically builds packages - correctly - and puts them into a repository for serving for the distribution releases we support. Salt Stack could have their CI pipelines do the same, using pbuilder, chroots, Docker containers or what have you. Then just have those development repositories available for people who want to test the bleeding edge.

This way, as soon as something no longer builds because someone used a feature not supported by X, you'll know almost immediately (probably it failed to even package), and can then make an informed decision as to whether or not you really want to use that feature. If you do, packaging of the dependency can proceed immediately where required - instead of waiting until it's decided that a feature release needs to be cut and scrambling to get it done.

dmurphy18:
Hi Adam

With regard to a section in the motivation part of your response:

Salt Stack could have their CI pipelines do the same, using pbuilder,
chroots, Docker containers or what have you. Then just have those
development repositories available for people who want to test the bleeding
edge.


SaltStack actually uses pbuilder, chroots, etc to perform building and has
done nightly builds to catch just the issues you raise.  However the change
which caused the issue with dependencies was due to having to switch to
using setuptools in Salt itself. The issue was caught quickly thereafter,
but it was a last minute change before release to fix an issue found in
final testing which had to be fixed.  So CI/CD issues would be the same, in
that the issue was detected quickly, fix worked on, and updates made, but
it was right before a release which necessitated the extra effort to ensure
timely delivery to dates stated.

Regardless of CI/CD architecture used, the problems would be the same of
finding and fixing issues in final testing close to release dates.

Will leave the points raised on Security, Low-End Hardware and Downstream
Packaging to be addressed by others.

But I will point out that with Tiamat it has enabled Salt support on older
platforms, for example: Fedora Core 18 32-bit based switchgear with limited
memory and disk space and no locale support, which would not be possible
using conventional packaging.  Not all platforms come with a supported
version of Python 3, for example Rasbian 8 on a Pi, but Tiamat would allow
for that eventuality if so desired.  Most embedded platforms today if
capable of running Python should have sufficient resources to support Salt
using Tiamat.

David Murphy
Unix Engineer
SaltStack
PO Box 876
Lehi, UT 84043, USA

office 801-207-7440
dmurphy@saltstack.com |  www.saltstack.com

ScriptAutomate:
A question that comes up for me on Tiamat being used for `salt` (or really any other project), and on bundling dependencies together as part of distribution, is open source licenses for these dependencies. Once `salt` is actually including the dependencies, we would need to ensure we have automated the updating of something like [salt/DEPENDENCIES.md](https://github.com/saltstack/salt/blob/master/DEPENDENCIES.md) and have it included somewhere, at a minimum, if I understand correctly?

What is planning to be done to ensure usage of Tiamat is within compliance of licensing requirements and conditions required of dependencies?

At a glance, these different licenses appear in the current version of the `DEPENDENCIES.md` file:

- [MIT](https://tldrlegal.com/license/mit-license)
- [MIT/X11](https://tldrlegal.com/license/x11-license)
- [GNU Lesser General Public License, V2.1](https://tldrlegal.com/license/gnu-lesser-general-public-license-v2.1-(lgpl-2.1))
- [GNU General Public License, V2.0](https://tldrlegal.com/license/gnu-general-public-license-v2)
- [GNU General Public License, V3.0](https://tldrlegal.com/license/gnu-general-public-license-v3-(gpl-3))
- [Python License](https://tldrlegal.com/license/python-license-2.0)
- [Apache License, V2.0](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0))
- [ZPL 2.1](https://tldrlegal.com/license/zope-public-license-2.1-(zpl-2.1))
- [Mozilla Public License, V2.0](https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2))
- BSD
- Public Domain
- ASL

I'm asking because I'm definitely not a legal expert, and I'm not sure how these kinds of things are normally addressed in instances of distribution like this. Normally if a system is required to pull down dependencies from something like PyPi, then the approach would differ because the problem of distribution is left to the user, which Tiamat would be changing, correct?

waynew:
IANAL but I would assume that given the Apache license of Salt, and the fact that we release the source in addition to binaries, means there's no shenanigans going on, but I could be wrong!

OrangeDog:
@waynew for example, bundling GPL dependencies may require salt to be licensed as GPL instead. Needs a legal check.

rongshengfang:
Can someone address @cachedout and @boltronics ‘ questions? These are important questions and I for one would love to know the answers. Thanks!

baby-gnu:
Hello.

> It would be a learning curve for people to understand how to add python libraries to the new environment.
> Adding python libraries to the new environment might be harder in an air gapped network.

As @OrangeDog noted, formulas are impacted since some of them require to install python libraries to enable some salt feature, for example in [libvirt](https://github.com/saltstack-formulas/libvirt-formula/blob/master/libvirt/python/install.sls).

One idea is to create a dedicated function specific to installing salt dependencies, for example:

``` sls
libvirt-python-for-salt:
  saltdep.installed:
    - name: {{ package }}
```

Then, the `saltdep` will:

- use `pkg.installed` if salt is installed from the distribution
- use `pip.installed` if salt is installed from pip
- use whatever is required to install the dependency in whatever situation

This could even solve the case where a user [install the minion using pip](https://github.com/saltstack-formulas/libvirt-formula/issues/76).

Regards.

OrangeDog:
@baby-gnu that's more complicated than you'd like, because different distributions have different packages available, and under different names. Fomulae already deal with these differences themselves, but Salt probably can't maintain an exhaustive python -> distribution package mapping.

baby-gnu:
> @baby-gnu that's more complicated than you'd like, because different distributions have different packages available, and under different names. Fomulae already deal with these differences themselves, but Salt probably can't maintain an exhaustive python -> distribution package mapping.

I think the mapping should be done by the formula as it's already the case. The idea is just to make an indirection on the function call.

I think that tiamat minion should export a grain to be used as a package name selection by the formula, in libvirt for example:

- default: `libvirt-python3`
- os_family:
  - Debian: `python3-libvirt`
  - Suse: `python3-libvirt-python`
- os:
  - CentOS: `python3-libvirt`
  - Fedora: `python3-libvirt`
- tiamat: `whatever-it-is-called`
- pip: `libvirt-python`

sagetherage (approved):

sagetherage:
> Can someone address @cachedout and @boltronics ‘ questions? These are important questions and I for one would love to know the answers. Thanks!

I am working getting solid answers. I apologize it is taking some time; I want to ensure we have the correct information, here and not simply making idle commentary.

sagetherage:
This SEP was an agenda topic at the Open Hour on 2020-SEP-10 you can [see the meeting notes here](https://github.com/saltstack/community/wiki/Open-Hour-Agendas-and-Notes:-2020-09#sep-26-package-salt-with-tiamat) and quoted below for quick reference.

_SaltStack Engineers Bryce Larson and David Murphy
Discussion begins at 18:58 on the video_

- Motivation    
   - When we make a new release, because all dependencies we package along with Salt, thre are so many that any time requirements change, it takes a lot of time to find source RPMs, etc. This holds us back from releasing more quickly.
   - The native Python environment for native minions is limited.
   - With Mac and Windows, we package dependencies all in one bundle and can create them within a couple hours. This makes us package Linux and other operating systems as easily. Then at release, users can simply tag and the PR is ready to go.
- In use, it's making it quicker and easier to build Salt on non-Windows and non-Mac operating systems. It's being used successfully with Arista currently. Previously, we had to support and backport a cumbersome amount of packages making maintenance more difficult.
    - Regular packages will contain Py files that can be modified. There will also be a single binary available that people can run for testing use (but not production use).
    - Packages are built more quickly when using Tiamat.
    - Potential downsides
       - We have to keep up on CVEs in the dependencies that we use. However, this is now quicker to turn around and we can address CVEs more quickly.
      - Packages without many dependencies will now be architecture-dependent when being installed.

cachedout:
Hi @sagetherage. Thanks for these notes and for the link to the video. I've had a look at them and watched the relevant section from the video as well but I'm afraid that my questions weren't addressed and I don't think I heard answers to the questions from @boltronics either. Could somebody please answer those? Thanks!

thatch45:
> > How (for example) are you going to get patches for undisclosed major CVEs, the way the distributions do?
> 
> This question by @OrangeDog is extremely important and so far as I can tell, didn't get a direct response. If I understand this proposal correctly, this has the potential to put Salt badly on the back foot if a bad CVE hits for a dependency. I understand that the desire is to respond quickly, but as @OrangeDog correctly points out, distros often have these prepared in advance while Salt will not. This is a potentially huge deal, IMHO, as it has the potential to expose systems to known flaws without any recourse but to wait for a new release of Salt and deploy it across large fleets of machines.
> 
> I'm looking here at what I assume would be a list of deps which would be bundled for an OS like Redhat: https://repo.saltstack.com/py3/redhat/7.8/x86_64/3001/ . Is that a reasonable way to determine which deps would be bundled? (It's not clear from reading this proposal.)
> 
> I picked python-requests to look at which is _just one dependency_ of Salt since it's both in the requirements files and in the repo I linked above and I presume would be bundled using this new method, correct? It has has 38 CVEs so far during 2020. In 2019, Redhat did at least 10 (and possibly more) updates to the requests package alone.
> 
> I see two paths here:
> 
> 1. You issue releases of Salt for every CVE across all your deps. This seems as if it could lead to dozens of a releases a year at a minimum. I struggle to see how this would be a win for Salt users and doesn't seem sustainable for the release team either.
> 2. You independently evaluate each CVE to determine whether or not the security risk posed introduces a security risk to Salt. This may be possible though challenging, but it would be good to learn more about how the team plans to achieve this.
> 

We will release updates for CVEs for deps, since we already have to do this for maintained deps I don't see how this increses our workload - As for requests, a dep with that many issues - requests is a disaster - should be removed from the codebase.

> So, my questions are as follows:
> 
> 1. Has anybody undertaken an audit to estimate how many times a year Salt would need to release new packages under this new model? Perhaps 2019 could be used as a way to make an estimation for future years?

Not formally, but we are familiar with the dep load because we do currently package all of these deps for multiple distros, tiamat will be easier to manage deps than how we are doing it now.

> 2. What would the policy for upstream CVEs be? Would a new release of Salt be released for every CVE discovered in an dependency or would they be independently evaluated to determine if they pose a risk to Salt itself? If the latter, could we discuss more about how this evaluation would happen and how such an evaluation could be done in a timely manner given that no advance notice by the upstream maintainers would be available to the release team?

When dep CVEs are discovered we will repackage a stable release of Salt with the updated dep to cover the patch

> 3. Which releases of Salt would be patched in cases where a security release of Salt is issued due to an upstream CVE? Would it just be the most recent version or would it go back a number of versions?

The tiamat packages still distribute the Salt code in source format and will be able to be directly patched, we will maintain our current CVE patching policy to release packages and patches according to the support schedule.
> 
> Thanks!

thatch45:
> A question that comes up for me on Tiamat being used for `salt` (or really any other project), and on bundling dependencies together as part of distribution, is open source licenses for these dependencies. Once `salt` is actually including the dependencies, we would need to ensure we have automated the updating of something like [salt/DEPENDENCIES.md](https://github.com/saltstack/salt/blob/master/DEPENDENCIES.md) and have it included somewhere, at a minimum, if I understand correctly?
> 
> What is planning to be done to ensure usage of Tiamat is within compliance of licensing requirements and conditions required of dependencies?
> 
> At a glance, these different licenses appear in the current version of the `DEPENDENCIES.md` file:
> 
> * [MIT](https://tldrlegal.com/license/mit-license)
> * [MIT/X11](https://tldrlegal.com/license/x11-license)
> * [GNU Lesser General Public License, V2.1](https://tldrlegal.com/license/gnu-lesser-general-public-license-v2.1-(lgpl-2.1))
> * [GNU General Public License, V2.0](https://tldrlegal.com/license/gnu-general-public-license-v2)
> * [GNU General Public License, V3.0](https://tldrlegal.com/license/gnu-general-public-license-v3-(gpl-3))
> * [Python License](https://tldrlegal.com/license/python-license-2.0)
> * [Apache License, V2.0](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0))
> * [ZPL 2.1](https://tldrlegal.com/license/zope-public-license-2.1-(zpl-2.1))
> * [Mozilla Public License, V2.0](https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2))
> * BSD
> * Public Domain
> * ASL
> 
> I'm asking because I'm definitely not a legal expert, and I'm not sure how these kinds of things are normally addressed in instances of distribution like this. Normally if a system is required to pull down dependencies from something like PyPi, then the approach would differ because the problem of distribution is left to the user, which Tiamat would be changing, correct?

Thank you, we have consulted with legal experts to make sure that the deps are in adherence to licenses. Salt already makes the information about deps available and we have made considerations for LGPL code as well. All deps are distributed in source code format and they are modifiable, information about them is available inside the distributed build and that code can be modified, either directly or using the pip extentions we are building into the tiamat build.

OrangeDog:
> All deps are distributed in source code format

So there isn't a "single binary" option?

thatch45:
There is a single binary dep, the contents of the binary are dynamically unpacked and executed. Tiamat can include third party shared libs and executables and can still access system libs and executables.

OrangeDog:
If you ship a binary that includes GPL code, then the whole thing has to be GPL.
Have you avoided doing this, or have you found a sneaky loophole?

thatch45:
The code is not actually statically linked within the binary. Within the binary the code is in a zip style archive. Since it does not statically link in the binary the distribution mechanism can be legally classified more like a dynamic installer. So yes, loopholing is happening :)

OrangeDog:
There are still some questions to answer here https://github.com/saltstack/salt-enhancement-proposals/pull/34#issuecomment-690371564

aplanas:
For reference, I am an openSUSE contributor and a SUSE employee.  This
fact will explain a bit some of my comments.

For context I want to mention some other data that can help me to
introduce later some other ideas:

  * The openSUSE Uyuni project (and the SUSE SUMA product) uses Salt
    too control, update and manage remote nodes.  Those nodes can use
    recent distributions (SLE15, RHEL8), some good old ones (maybe
    SLES11 or RHEL5), and some other not very widely used (and that is
    OK, because Salt is very good doing the best effort is many cases)

  * Uyuni (and SUMA) are multiarch: Intel, Aarch64, S390X, PPC64LE

  * openSUSE uses a build service to create packages for openSUSE and
    other distributions, and we are packaging Salt ourselves:
    https://build.opensuse.org/project/show/systemsmanagement:saltstack:products
    (see also the subprojects)

  * We take care of the dependencies, and we (as today) use the Python
    installed system.

For those who does not know OBS, is a community (meaning open to
anyone) build service that take the package specification file, the
source code (in tarball usually, but can be also a git repository that
is excellent for CI/CD) and some patches and build matrix (target OS
version X target architecture).  OBS will create VMs for those systems
and will build inside the package, so will generate (is succeed) from
the *same* set of source assets one RPM, DEB, or whatever for each
element of the build matrix.

The big deal is OBS will detect that we change any element of the
dependency tree and will rebuild a version for us for those systems
that needs it.  This is totally important when we fix a CVE in one
package used directly or indirectly by Salt.

Knowing that, we also share some of the problems that this PEP 26 is
addressing: (a) old Python versions installed in the system and (b)
old Python modules that comes from repositories out of our control.

We would like to solve the same issue, and this is why I want to share
this idea: evaluate OBS as a build system to generate the single
executable that is described in PEP 26.

Ideally Salt will have its own project in OBS, with full control
(maybe open to collaboration of other users via submit requests, that
is a similar concept that a pull request), and in-there will generate
the Salt official executable package.

There are adjustments tho, like for example that Tiamant an PIP are
deeply problematic.  PIP being the real big issue, and download from
PyPI packages that are later delivered to customers, and that cannot
be patched (fixed) easily.

We resolved that.

Sometime ago we did this modest proof of concept, following the
leadership of Tiamat. Is living here:

https://build.opensuse.org/package/show/systemsmanagement:yomi/salt-bin

Internally we are not using Tiamant, virtualenv nor PIP to achieve the
same goal.  We are using PyInstaller and the entry point (run.py) that
belongs to Tiamat, and with that we can generate one executable per
each combination of distribution and architecture.

The spec file that drives it is only 61 lines, including comments and
license. The niceties of this approach are:

  * Totally reproducible and signed builds

  * A fix in one dependency will generate a new Salt binary that
    contains the fix, automatically

  * Includes the same Python interpreter and Python dependencies, and
    you can control both from the spec file

  * You can decide what libraries from the target system you will use,
    and what libraries to include in the final binary. This can fix
    the libvirt issue described here, that is also something impacting
    use greatly.

  * You can try to create a single binary that will be used in
    multiple different distribution (for the same architecture), and
    specific binaries for outliers (maybe because use a super old
    glibc)

  * There is a team of openSUSE / SUSE engineers that can help

I understand that using OBS have also drawback and incompatibilities
with internal workflows and policies, that can draw it as an
unfeasible approach.

We are exploring this single binary solution, and a different one
based on a helper script (https://github.com/SUSE-Cloud/venvjail) to,
at the end, resolves the same problem (decoupling from system's
provided Python, modules, and libraries, together with easy patching
and multi-architecture), and we are still in the early research stage.

We still do not have all the data to make an informed decision, we are
on it.  But in the case that Tiamat and PIP becomes the official Salt
distribution, we should discuss how it will impact other distribution
mechanism, that the one that openSUSE / SUSE is developing.

For example, we can find hidden tricks in how salt-ssh create, select
and deploy the binary asset.  IMHO Salt should technically leave the
door open to a different mechanism, maybe blessing those alternatives.

Anyways, thank you for coming to my ted talk :-) I truly think that
this is an interesting PEP.

boltronics:
> We will release updates for CVEs for deps, since we already have to do this for maintained deps I don't see how this increses our workload - As for requests, a dep with that many issues - requests is a disaster - should be removed from the codebase.

I assume this was also stated to partly answer the first part of my post above.

From the Python [docs](https://docs.python.org/3/library/http.client.html):
> The Requests package is recommended for a higher-level HTTP client interface.

Would we really want to remove this? Would Salt be able to avoid the mistakes that a dedicated library has made? (I haven't looked into what the issues were.) Would we remove this dependency (and any other with a high security fix rate) before adopting Tiamat?

OrangeDog:
> a dep with that many issues

Usually, having frequent security fixes to a widely-used library is a good thing. Newer or more obscure libraries are likely to have the same critical flaws, except they haven't been fixed yet.

aplanas:
w.r.t CVE. Is my understanding that with Tiamat now we needs to address all the CVEs of the dependency graph closure, not only the ones that affect directly Salt?

If that is the case I can confirm that this is a good amount of work for the release manager.

I really advise to check this OBS build system, as resolve the CVE problem for free for the dependencies. SaltStack should only needs to provide the CVEs patches for the Salt code itself, like already is doing (that is also work)

ScriptAutomate:
A basic, minimal readme has been added to `salt-pkg` on GitLab, with directions on how to install the beta builds of `salt` built from Tiamat:
- https://gitlab.com/saltstack/open/salt-pkg/-/blob/master/README.md

That repository is dedicated to the packaging of salt.

sagetherage:
## [Meeting Notes from 2020-OCT-15](https://github.com/saltstack/community/wiki/Open-Hour-Agendas-and-Notes:-2020-10#sep-26)
- _Conversation during the meeting_ 2 different items or things here
   - extending salt to load more modules into a salt environment to pick it up - recently available in v7.3
   - pip install into the system - not yet addressed as of these notes ([Comment #34](https://github.com/saltstack/salt-enhancement-proposals/pull/34#issuecomment-690371564)) need to fix on the salt side, default to the python environment on the system and will be addressed for the Magnesium release or the following point release (**correction** _at the time of the original notes the Core team attempted to correct for the Mg release, and more work will continue for the Mg point release in November_)
      - Merged as of 2020-OCT-21:
      - GitLab saltstack/pop/taimat-pip MR#4 [Tiamat-pip now keeps a record of packages](https://gitlab.com/saltstack/pop/tiamat-pip/-/merge_requests/4)
      - GitLab saltstack/open/salt-pkg MR#50 [Bump tiamat-pip requirement to 1.1.0](https://gitlab.com/saltstack/open/salt-pkg/-/merge_requests/50)
- [README](https://gitlab.com/saltstack/open/salt-pkg/-/blob/master/README.md) for beta tiamat packages of Salt

## [Meeting Noes from 2020-OCT-22](https://github.com/saltstack/community/wiki/Open-Hour-Agendas-and-Notes:-2020-10#tiamat-and-salt-sep)

"Because Tiamat packages its own version of isolated Python and pip, it means that the functionality around pip packages install calls from Salt will have changed default behavior. An execution or state module call to install Python packages from pip would need to specify the system-level Python/pip if it wants to install packages that become available to system-level/third-party Python tools and applications outside of salt.

Pedro will be responding to questions, comments, and concerns around this in the SEP on Tiamat packaging for Salt."

s0undt3ch:
> Will minor Salt upgrades require reinstallation of 3rd-party python libraries? Any other potential upgrade issues?

No. Python doesn't break it's own compatibility with minor upgrades.
If however, the new salt tiamat packages upgrades the Python major release, it's safe to say that one _should_ re-install 3rd-party libraries. I'll expand more on this at the end.

> This issue affects regular users, formulas, and custom modules.
>
>Many modules require an additional python dependency in order to work.
>You can get it with either pkg.installed or pip.installed (which may also require pkg.installed for dev dependencies).
>
>How is this going to work now?

Right now, exactly as you stated above, using `pip.installed`. More on this at the end of this message.

>Are all the formulas going to need to detect whether this is a tiamat installation in order to do the right thing?

No. Not currently

>Is the default environment for pip.installed going to be the system python, or the tiamat one (either way requires people to change something)?

On the first beta tiamat packages of salt, the default environment will be tiamat's

>Will system python packages be available in the tiamat environment?

No.

>What about compatibility between non-python system packages and those bundled by tiamat?

This is where targeted distro packages of tiamat come in. We currently build for the RHEL and Debian families, on each of the supported versions, so, this concern should be nonexistent.

> This question by @OrangeDog is extremely important and so far as I can tell, didn't get a direct response. If I understand this proposal correctly, this has the potential to put Salt badly on the back foot if a bad CVE hits for a dependency. I understand that the desire is to respond quickly, but as @OrangeDog correctly points out, distros often have these prepared in advance while Salt will not. This is a potentially huge deal, IMHO, as it has the potential to expose systems to known flaws without any recourse but to wait for a new release of Salt and deploy it across large fleets of machines.

If this is a python-dependency, users have the options t `salt pip install -U <said-python-dependency>` before a new tiamat package is made available. But yes, our tiamat packaging needs to be on top of these issues.

> I'm looking here at what I assume would be a list of deps which would be bundled for an OS like Redhat: https://repo.saltstack.com/py3/redhat/7.8/x86_64/3001/ . Is that a reasonable way to determine which deps would be bundled? (It's not clear from reading this proposal.)

Tiamat builds will make use of the packaging static requirements, the full bundled python libraries are `requirements/static/pkg/py3.*/linux.txt`


### Tiamat pip support update

The [tiamat-pip](https://gitlab.com/saltstack/pop/tiamat-pip/) project was created to address extending the python packages available to salt.
You can install and uninstall any 3rd-party python libraries you need, **even if it has to link to the "embedded" python interpreter or a system library**.
You cannot, of course, uninstall a bundled python dependency, but, theoretically(need to add a test case), you can upgrade it.
Tiamat-pip keeps a record of what packages are installed/uninstalled. The purpose of this functionaity is to support rebuilding/reinstalling the 3rd-party libraries in case the tiamat package upgrades the bundled python major release version. (For now only the recording is implemented. An easy way to reinstall is not yet implemented).

**NOTE: Salt's `pip.installed` will default to tiamat's python environment.**
What this means is, if your current state tree only uses `pip.installed` to extend the salt support, you're good and nothing needs to change.
If, on the other hand, your state tree is supposed to extend the system python(first, you should be using system packages :grin:), then you need to be explicit about the `bin_env` parameter to `pip.installed` and point it to the system python binary you want to use.

This opens up a question, which python environment should `pip.install` default to?
[](https://api.gh-polls.com/poll/01ENK1F2TKK08SDJS7A7M452WW/%60pip.installed%60%20should%20default%20to%20the%20system%20python/vote)
[](https://api.gh-polls.com/poll/01ENK1F2TKK08SDJS7A7M452WW/%60pip.installed%60%20should%20default%20to%20bundled%20python/vote)

OrangeDog:
Is tiamat going to bundle non-python dependencies or not? For example libgit2 and libzmq.

None of the answers have made reference to them either way. If they're not bundled then I don't see how the dependency/testing problems are solved, and if they are then this need to talk about how they're synchronised with the python dependencies, upgraded, or new ones added, and whether the licensing is valid.

s0undt3ch:
> Is tiamat going to bundle non-python dependencies or not? For example libgit2 and libzmq.

Yes and No.
No, the python libraries that require those will link to them at install time. PyZMQ ships it's own linked zmq bindings on its wheel.

As for libgit2, we're currently not including it in the tiamat package. If we were and linking was not enough, ie, it actually needed libgit2 available, we would include it in the tiamat package.
Since we're not including pygit2, you'd have to install libgit2 on your system and then pip install pygit2, which either links or loads libgit2 at runtime.

> If they're not bundled then I don't see how the dependency/testing problems are solved,

How come? I don't understand the problem you're trying to point out.

> and if they are then this need to talk about how they're upgraded, or new ones added, and whether the licensing is valid.

@thatch45 has already addressed the licensing concerns. Are there any more concerns regarding licensing?

As for how and when they are upgraded, at the lack of a written plan, at least so far, I'd say we upgrade to latest releases every time we get out of code freeze to start working on the new release, or if/when needed because of a security release.

sagetherage:
Meeting notes from the Open Hour on 2020-NOV-19 discussion and communication of the poll in the above comment can be seen here: https://github.com/saltstack/community/wiki/Open-Hour-Agendas-and-Notes:-2020-11#existing-seps

s0undt3ch (approved):

aplanas:
Just an update here from https://github.com/saltstack/salt-enhancement-proposals/pull/34#issuecomment-710107936. We have a very rough and broken prototype of a bundled version of Salt, as described before.

The source code is living here:
>  https://build.opensuse.org/project/show/systemsmanagement:saltstack:bundle

The targets for the different OSs are living in subprojects:
> https://build.opensuse.org/project/subprojects/systemsmanagement:saltstack:bundle

For example, this is for CentOS:
> https://build.opensuse.org/project/show/systemsmanagement:saltstack:bundle:CentOS-7

The only required package is `venv-salt-minion`, the rest are build dependencies
> https://build.opensuse.org/package/show/systemsmanagement:saltstack:bundle/venv-salt-minion

Some notes of this exercise:
* Is cheap. Is a couple of weeks we bundled Salt from the code base of SLE15-SP3. The amount of patches required to target different distros is not very high!
* Debian was a problem, but extending `debbuild` to support `lua` was critical to support the RPM macros.
* Fix of CVE for dependencies comes for free! For example, meanwhile I was writing this comment, the repository received a fix for the CVE-2021-3177, without any involvement from any of the Salt maintainers.

There is another nice technical thing, that I do not know how Tiamat can address. There is this subtle problem with Python bindings that comes from system libraries.

For example `python-rpm`. CentOS / RHEL / SUSE uses this RPM package format, that provides Python bindings that are written in C. This binding connect in one side the `librpm` part from the system, and for the other the `libpython`. Because RPM, depending on the version, can store the `rpmdb` in different places (`/usr`, `/var`) and with different formats (`ndb`, `bdb`, `sqlite`), is really important to use the same `librpm` version that the system expect (you can break your system otherwise).

We addressed this problem in OBS importing the pristine `src.rpm` from the target system, and compiling it with the Python version that will live inside the bundled (that is totally different from the one in the system). This generate a new python-rpm that is included in the bundle, with the correct library links from both sides (system and internal Python). For example, the one for CentOS is here:
> https://build.opensuse.org/package/view_file/systemsmanagement:saltstack:bundle:CentOS-7/rpm/_service?expand=1

waynew (approved):

cmcmarrow (approved):

twangboy (approved):

garethgreenaway:
Just wanted to give everyone a heads up on some recent changes you may or may not have seen related to the Salt bootstrap script or the Salt 3005 documentation.  The change in question was to rename the wording around the new onedir packages that are being built for 3005 using the Tiamat tool.  The core team discussed this topic at length and decided that it was more accurate and descriptive to refer to these packages as onedir rather than referring them to as tiamatpackages.   This gives us a clear distinction between the new onedir packages and the classic packages by focusing on the concept used to build them rather than the tool used to build them.

PR #40, opened 2020-12-21
Aadding the ability to add beacons automatically from the state system

garethgreenaway:
Adding a SEP for change proposed changes to state system, adding the ability to add beacons automatically from the state system.

### DIFF
--- a/add_beacons_from_states.md
+++ b/add_beacons_from_states.md
@@ -1,0 +1,41 @@
+- Feature Name: Add Beacons from Salt States
+- Start Date: 2020-12-21
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Adding a new global option to Salt state system that would audomatically add a beacon on the targeted minion for specific functions in specific state modules.
+
+# Motivation
+[motivation]: #motivation
+
+Adding this new option would allow beacons easier and associating them with Salt managed resources.
+
+# Design
+[design]: #detailed-design
+
+The design and implementation of this new option would done in a similar way to the mod_watch system.  During a state run
+the various individual states will be scanned and those that include the beacon option will be run through a function that
+calls a "mod_beacon" function in the desired state module.  This "mod_beacon" function would create the beacon using the
+configured resources associated with the state in question, eg. the PKGs, files, or services.
+
+## Alternatives
+[alternatives]: #alternatives
+
+N/A
+
+## Unresolved questions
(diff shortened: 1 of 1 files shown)


Akm0d (approved):

max-arnold:
To be honest, right now, this looks like a limited use-case that was requested to be bolted-on without considering the broader context and long-term implications. It seems limited because a global state option is linked to a narrow function (beacons). And the broader context is that the ability to run arbitrary functions in response to particular states is much more powerful and flexible (and it is a more substantial reason to complicate the Salt state system).

Also, it is not entirely clear what specific use-cases require this change (the motivation sentence is quite vague: `Adding this new option would allow beacons easier and associating them with Salt managed resources`).

Have you considered a more generic implementation like the [transparent requisites](https://gitlab.com/saltstack/pop/idem/-/blob/master/docs/topics/transparent_req.rst) in Idem (or Salt executors, although they are much more limited)?

What I'm thinking about is having something like pre/post/onchange/onfail hooks that can be assigned to particular states globally (with the ability of more granular filtering based on state arguments and also enable/disable on a per-state basis). For example, it could possibly allow refactoring of the hardcoded [file backup option](https://docs.saltstack.com/en/latest/ref/states/backup_mode.html#file-state-backups) into a hook (a global requisite).

Another critical point is that it should be possible to place these global hooks (requisites) into arbitrary (custom) modules and not just into the same module where the state resides. This will enable the development of custom global policies and automations for Salt states without altering existing modules and state files.

OrangeDog:
Could the same thing be achieved with a state that manages beacons, and the current available requisites?

sagetherage:
This SEP does need some updates and phase 1 was implemented in Aluminium, prior to this being approved. That isn't the process and we made a mistake, here. @garethgreenaway between now and the next 2 weeks can you please make adjustments to this SEP and address the comments, please? 

Here is the PR of Phase I https://github.com/saltstack/salt/pull/59559

waynew:

---

Summary of the current discussion:

There are questions as to whether or not this could be accomplished in another way, or if this could be improved by taking a more holistic view of the desired functionality.

garethgreenaway:
The idea was to go beyond that is done with states and requisites, per the example in the SEP...you use file.managed to create a file, a running beacon would be added to the minion that watches that file for changes, and then fires an event, which runs the original state (or part of it) to put the file back.  Obviously you can do all this manually setting up you're own reactors or using the beacon state module to create a beacon.  Considering this could be a common use case, the idea here is to do it automatically within the state system without requiring the user to worry about creating the beacon state to manage the beacon or worry about having the proper requisites in place.

PR #41, opened 2021-01-19
SEP 31: Allow Salt to be installed anywhere on Windows

twangboy:
This is a SEP to allow Salt to be installed anywhere on Windows. The default would be to `C:\Program Files` instead of `C:\salt`


windows-install-anywhere.md
```diff
@@ -0,0 +1,108 @@
...
+
+There is some discussion to be had over the name of the <salt> directory. Here
+are a few proposals:
+
+- Salt
+- SaltProject
+- Salt Project **
+- saltproject.io
+
+We recommend ``Salt Project``
```
s0undt3ch, line 67: Are spaces in directory names on windows a common source of problems?

OrangeDog: Yes, for poorly-written software. Having spaces by default might be good for uncovering any bugs that get be fixed.

xeacott: Yeah our **number 1 goal** with this SEP is to create as much consistency as possible based off typical installations of software for Windows/MS.

OrangeDog: Note that "Program Files" already has a space in it, so another space in "Salt Project" isn't going to make any difference.

s0undt3ch: Just trying to raise awareness.

@orangedog, `c:/salt` doesn't have spaces, that's the current default.

Of course, if you add one space to the path, 2 or 3 more won't hurt :)

waynew: Salt *shouldn't* have any issues with being installed in a different location, since we should be using all of the builtin Python pathlib/os.path functions.

cruscio:
Please consider setting defaults using environment variables like `%PROGRAMDATA%` and `%PROGRAMFILES%`, rather than their common (but not definite) values. I've seen more than a few clients with no `C:` drives on their servers.

amalaguti:
Make sure it can be installed in other than C drive.

dafyddj:
What is the plan for Minion upgrades from the current setup (`C:\salt` "old-style")?
I can see 3 scenarios and wonder whether the installer will allow for one, some or all three?

1. Upgrade leaving the old-style directory structure in place.
2. A mandatory change to the new default directory structure (`C:Program Files\Salt Project`).
3. Allow an optional change to the new default directory structure.

I guess the ideal would be to allow for all three scenarios, but I don't know whether that's practical.

dafyddj:
How will directory permissions change? At the moment `C:\salt` is locked down to `Administrators\SYSTEM` only, but this won't be suitable for the new-style directory structure.

Presumably the binaries will be readable by all, but what about the other directories?

OrangeDog:
> this won't be suitable for the new-style directory structure

Why not?

garethgreenaway (approved):

s0undt3ch (approved):

dafyddj:
> > this won't be suitable for the new-style directory structure
> 
> @dafyddj Why not?

The default permissions for directories under `C:\Program Files\` is Full Control for `BUILTIN\Administrators` and Read/Execute permissions for `BUILTIN\Users`, and I don't see any reason why we wouldn't want to go along with that, particularly as it feeds into one of the reasons for making this change, which is enabling running Salt as accounts other than `SYSTEM`.
I'm just trying to prompt discussion for what to be done with the other "Program Data" type of directories.

cruscio:
Non-administrator users should not, by default, be allowed read permissions to anything that may contain sensitive information. pki, cache, job history, log files, etc.

0031-windows-install-anywhere.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Windows is not Linux
+
+The Windows directory structure is such that installed software should reside in
+the ``Program Files`` directory on the ``C:`` drive. Software should not be
+installed on the root of the system directory. Salt is currently installed on
+the root of the ``C:`` drive. This seems to be a paradigm brought over from
```
OrangeDog, line 21: > This seems to be a paradigm brought over from Linux.

What does this mean. I have never, ever seen anything installed directly at the root on Linux. The closest you get is people dumping stuff under `/opt` and not using `/etc/opt` or `/var/opt`.

xeacott: Perhaps this was overstated and not well written out. What we wanted to point out from this specific paragraph was just simply that we have the following:
```
C:\salt
C:\salt\var
C:\salt\bin
C:\salt\srv
```
and this was something that was carried over from linux whereas this convention is just simply not correct for Windows. Therefore, we wish to remove prior linux conventions.

xeacott:
> Non-administrator users should not, by default, be allowed read permissions to anything that may contain sensitive information. pki, cache, job history, log files, etc.

Precisely. Which, was brought up by @dafyddj today in the Window's windows group. This can especially be dangerous as, this in theory would allow for a non-admin account to have access to Salt, and allow a custom module to be synced causing non-desirable behavior. So where ever these sensitive files resides, we'll lock them down in the exact same way.

damon-atkins:
Just want to point out their are a few Microsoft products which do not install everything in C:\Program Files  some times C:\ not large enough.  Some companies prefer to keep C:\ clean and install on D:\

xeacott:
> Just want to point out their are a few Microsoft products which do not install everything in C:\Program Files some times C:\ not large enough. Some companies prefer to keep C:\ clean and install on D:\

Yep! Been addressed above: 
```
There are scenarios where the System drive contains only the system files with
all other software installed on a separate drive, D:\Program Files for
example. This is done by setting the environment variable ProgramFiles. The
default location for program data can also be set using ProgramData. It is
proposed that Salt uses these environment variables to determine the default
location of Program Files and ProgramData
```

xeacott (approved):

krionbsd (approved):

bryceml (approved):

sagetherage (approved):

PR #48, opened 2021-05-30
SEP 32: master key rotation for disconnected minions

marbx:


master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,107 @@
...
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+There are several reasons to rotate keys:
+- Exposure or a compromised key
+- Security departments require key rotation on a periodic basis
+- It is common praxis to expire or revoke keys or certificates (e.g. [eset protect 1](https://help.eset.com/protect_admin/80/en-US/certificate_replacement.html), [eset protect 2](https://support.eset.com/en/kb6824-peer-certificate-is-going-to-expire-error-in-eset-security-management-center-7x))
```
OrangeDog, line 13: practice

OrangeDog: 2 & 3 are really the same thing.

master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+If a minion is offline when the master changes its key, there is no way to get it after the fact. The minion will not be able to reconnect to the master after the master has changed its key. This is a problem e.g. with end user desktops/laptops that aren't ALWAYS connected to the master.
+
+The goal is a key management system that allows for:
+- Expiring and future (post-expire) keys
```
OrangeDog, line 22: Expiring and ~~expired~~ replacement keys

master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,107 @@
...
+- Disconnect minion after key1 expired and reconnect
+
+Can it be automated?
+ - Yes, this is a highly automatable feature
+
+
+## Alternatives
+[alternatives]: #alternatives
+
+What other designs have been considered?
```
OrangeDog, line 74: The other obvious solution is to support multiple keys at once.
Then there's no need for any time-based stuff.

master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,107 @@
...
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Please consider:
+
+Implementation cost, both in term of code size and complexity
+- Each minion must additionally store the post-expire key and the expire time, which is below 1 KB
+- The key rotation action is essentially moving a file
+- The key rotation decision must be taken before re-authentication and is essentially comparing present time with expire-time
+Integration of this feature with other existing and planned features
+- The advise [ROTATE A MASTER KEY](https://docs.saltproject.io/en/latest/topics/hardening.html#rotate-a-master-key) contains a security hole because the minion must blindly trust the new key
```
OrangeDog, line 98: advice

OrangeDog (changes requested):

marbx:
Hi @twangboy thank you for your commit to the SEP32. Do I need to take an action regarding you commit?

master_key_rotation_for_disconnected_minions.md
marbx: Thank you, correcting both

master_key_rotation_for_disconnected_minions.md
marbx: Thank you for suggesting `replacement key` - this improves readability.

master_key_rotation_for_disconnected_minions.md
marbx: I also don't like the "time-based stuff" so much. 
Without "expire time", the minion could use the replacement key could once the current key no longer works. This should be the sign that the server has switched to the replacement key. But here still only one key is active at a time.

Could you please explain how to support multiple keys at once?

master_key_rotation_for_disconnected_minions.md
marbx: Thank you, corrected

master_key_rotation_for_disconnected_minions.md
marbx: I added the "time-based stuff-less"  alternative

OrangeDog: I'm not sure what needs explaining? If both keys work at the same time then there is no need for this complexity, Simply add a new master key, and then once all the minions are aware of it, remove the old one.

marbx: Right. I'll reduce "expire time" to an alternative.

master_key_rotation_for_disconnected_minions.md
OrangeDog: I meant the second and third bullet points are the same thing. It is common practice to rotate keys to limit future exposure.

dwoz:
I like the premise of this SEP and have some ideas around implementation. What I would like to see is the ability to customize how long keys active. I would also like to have the ability to pre-seed multiple keys. It would be nice to have a list of keys and date ranges those keys are valid for. This would allow users to customize how long of a window minions will have to get the new key. In some situations you may only want to allow a week for minions to get the new keys, others maybe a month. The following is pseudo a config:

```
keys:
   2021-08-01, 2021-09-01: key1.pem,
   2021-08-15, 2021-09-15: key2.pem,
   2021-09-01, 2021-10-01: key3.pem
```

With the above config keys would be active for 1 month and minions would have 15 days to get the new keys.

master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Definition of any new terminology
+  - Current key (k1): the key that is currently in use.
+  - Replacement key (k2): a key which is unused as long k1 works.
+  - Rotation: replacing k1 with k2.
+- Examples of how the feature is used
+  - Create replacement key k2.
+  - Distribute k2 to all minions, which could require to wait for offline minions,
+  - After complete distribution of k2, the master can rotate keys from k1 to k2.
+  - When a minion reconnects with k1 and fails, he reconnects with k2.
```
ITJamie, line 42: "he" should be replaced with "it"

marbx:
@dwoz, I am fine with specifying date ranges for keys.

Just for clarity: between 2021-08-15 and  2021-09-01, both key1 and key2 are valid?

ITJamie:
What about where there is only currently one key with no range set. In
future i decide to add another. But again i dont want to add a range.
Aka i just decide to do a key rotation and set some kind of max time on the
old key.

master_key_rotation_for_disconnected_minions.md
```diff
@@ -0,0 +1,81 @@
...
+  - No change on the master required.
+  - A minion must additionally store the replacement key.
+  - The key rotation action is essentially moving a file.
+- Integration of this feature with other existing and planned features
+  - The advice [ROTATE A MASTER KEY](https://docs.saltproject.io/en/latest/topics/hardening.html#rotate-a-master-key) contains a security hole because the minion must blindly trust the host that responds as `master`.
+  - There is no API to rotate keys, only a [rekey script](https://github.com/dwoz/salt-rekey/) that requires all minions to be connected at the same time.
+- Cost of migrating existing Salt setups (is it a breaking change?)
+  - No migration costs, not a breaking change.
+- Documentation (would Salt documentation need to be re-organized or altered?)
+  - Only added.
```
waynew, line 81: There are also security considerations that should be discussed and addressed: how do we audit the keys & range changes? Currently it's basically static, but we're adding the ability for an attacker to potentially add (a) new key(s). This is especially important for the `public_key=/full/path/to/file` - we need to ensure that the file is only writable(readable??) by root or whoever Salt is.

If the key is readable by non-Salt users that means that it's pretty trivial to MITM salt, and if it's writable well... same problems but even worse :upside_down_face:

I think generally I like the idea of this SEP, but this SEP should be more detailed than a typical SEP, especially around the security considerations. Given the area this touches we should get as many eyes on it as possible, including the security WG at *minimum*. We should get some VMware security review as well.

waynew (changes requested):

waynew:
The date range question is an interesting one. I *think* I like it, but we should clarify the gamut of potential cases:

- a single key with no range (current state)
- a single key with a range(???) (are the endpoints exclusive or inclusive? Do they roll over at midnight? What timezone? Always UTC? Whatever the master uses?)
- 1+N keys with a range, where there is overlap
- 1+N keys with a range with no overlap - do we warn? Error?
- Do keys always need a start/end? Can we use a MINIMUM constant or just recommend people use 1900-01-01 or something?

There are probably several other scenarios that I haven't considered, but those at least should be discussed/addressed in this SEP

master_key_rotation_for_disconnected_minions.md
marbx: @waynew wrote

> we need to ensure that the file is only writable(readable??) by root or whoever Salt is.

You need this already today and it is ensured by verify.py today.

marbx: @waynew wrote

> we're adding the ability for an attacker to - potentially add (a) new key(s)

That's not true. What you call "ability" requires access to the API.
Attackers who can access the Salt API can add a key today. The current API is powerful enough.


However, the API under discussion makes it easier to rotate keys. This increases security, as described.

marbx:
@ITJamie 
> What about where there is only currently one key with no range set. In
> future i decide to add another. But again i dont want to add a range.
> Aka i just decide to do a key rotation and set some kind of max time on the
> old key.
> 

This is also my use case.

master_key_rotation_for_disconnected_minions.md
waynew: IIUC (it's been a while since I've been in that code) - currently master key rotation is an out of band process, and requires the Salt master to restart even to enable a reload of the key. And like the highlander - there can be only one.

This isn't necessarily more or less secure (as that depends on your objectives). But currently *salt* doesn't really support changing out the key.

And if the key changes then minions will complain loudly.

But that's not the case if we change things to allow multiple keys. I'm not saying that this is a gaping security hole, but I'm saying that we should be mindful of the *possibilities*.

ITJamie:
> @ITJamie
> 
> > What about where there is only currently one key with no range set. In
> > future i decide to add another. But again i dont want to add a range.
> > Aka i just decide to do a key rotation and set some kind of max time on the
> > old key.
> 
> This is also my use case.

To me the goal would be to allow some kind of concurrency with both keys. 
aka if a client try's to connect with the "old" key the server side should accept it but then let the client know about the new key.

dwoz:
> @dwoz, I am fine with specifying date ranges for keys.
> 
> Just for clarity: between 2021-08-15 and 2021-09-01, both key1 and key2 are valid?

yes

waynew:

---

Summary of the current discussion:

There are a few feedback items that still need to be addressed (see review comments) - generally the discussion is favorable, but there are concerns about ensuring that attack surfaces aren't introduced.

barneysowood:
# Use cases

From the SEP and linked issues, it looks like there are two scenarios it's trying to cover:

1) Master is compromised and minions need to accept the new master's key
2) Master key is required to have a limited lifetime due to some policy

For 1, when the master is compromised and that is detected the SEP is aiming to have a process to allow minions to connect to a new master with a new key. In this case, the minions fall into two categories:

a) Minions that were connected to the master between the time of the compromise and the time of detection or remediation.
b) Minions that were not connected to the master in it's compromised state.

I'd argue that for a) there's not a lot this process can do ensure the key is safely updated on the minion - unless it's absolutely certain that the compromised master was not used to run any jobs on the minion *and* that the key wasn't extracted and used on another host impersonating the master that could have run jobs on the minion. I think you have to assume that the minion is compromised. I'm not sure trying to ensure it is safely connected to a new master makes sense.

For b) it may be possible to safely update the master key, but again only if you really are sure that the minion didn't connect to the master and didn't connect to something impersonating the master.

I think the main point is that any process that is implemented isn't going to guarantee that you can safely move minions from a compromised master to a new master.

For 2, it's not totally clear from the SEP why the a limited lifetime is required beyond "policy", it would be great if others could expand on this. I would guess it's at least partly to deal with undetected compromise of the master/key.

If attackers are able to extract the key but not maintain access to the master, limiting the lifetime of the key reduces their window of opportunity to use it to compromise minions. If they do maintain access to the master undetected, any system for rotation that uses the same master will not help. Maybe there is some other reason?

barneysowood:
## Implementation

The solutions discussed, both in terms of key distribution and things like keylifetime, seem to be implementing quite a lot of functionality that is common to existing PKI solutions. I'm wondering if it would be worth exploring the idea of having a minion trust a master based on certificate signed by a CA rather than it being a key it recognises? It might be possible to utilise existing and well tested libraries to add that functionality rather than rolling from scratch.

That would gain us quite a lot of nice properties:

* Minions only need to know/trust the CA - any master with a suitably signed cert would be accepted
* Minions could leverage OS or hardware certificate stores
* CRLs could be used to revoke master certs quickly
* CA keys would not be stored on the master
* Certificate lifetime could be set
* Software like vault that can generate very short lived certs could be used if required

It would also be a solution that organisations would find easier to understand and probably fit in to existing process for remediation and key handling.

I'm currently not sure how this would fit in to the existing master/minion communication. It might be possible to build this out of the normal path, perhaps if the minion connects an there's an unknown master, it asks for the new master pub key and cert and if it's signed by a trusted CA, installs the master pubkey and re-tries. It might actually be possible to build a simple prototype:

- Master has a webserver running on known port using TLS and a cert signed by trusted CA
- Webserver just serves up the master pub key
- Simple module on the Minion that knows how to connect to the known port, validate the cert and install the new key.
- The module could be called if the minion fails to connect on startup
- For improved security, the webserver could require TLS auth from the client.

## Using existing existing functionality

There may be another option that can get most of the way with only minor changes to existing code. That would be using the [multi-master PKI setup](https://docs.saltproject.io/en/latest/topics/tutorials/multimaster_pki.html), but not necessarily using more than once master at once.

So on the master you would set `master_sign_pubkey: True` and deploy the master_sign.pub pubkey at deployment time. You wouldn't (necessarily) configure multiple master in the minion config.

Critically, what you would do is use `master_use_pubkey_signature: True` and store the master_sign.pem key offline (or even better generate the key and signature offline) so that if the master is compromised you do not have to rotate the master_sign key. Any new master with it's pubkey signed with the master_sign key would be accepted by the existing minions.

The only additional code or process that would be required, would be a way to mark the old master key as revoked. Ideally there'd be a way to sign the revocation with the master_sign key and push it to the minion, but as an initial implementation a config option as a list of revoked master keys (hashes of maybe?) could be added and checked in salt/crypt.py. That way all the new master needs to do is push that config change to any minions.

This doesn't deal with time limited keys etc, but does seem like a reasonably simple way of getting disconnected minions to trust a new master.

PR #49, opened 2021-06-01
SEP 33: Re-adding point zero to major Salt versions - process update

sagetherage:
Adding a draft SEP to re-add the point zero to major Salt versions as a standard operating procedure or to the process. 
- Intent is not to adjust the past, but to explicitly state the process going forward and why
- Specific tasks and adjustments will need to be made, this is to discuss, plan, and create the premise to implement; meaning we agree the process and standard should be in the format of `####.#` example: `3005.0` and once approved the date at which this will become the standard will be announced/communicated

### DIFF
--- a/0000-readd-point-zero.md
+++ b/0000-readd-point-zero.md
@@ -1,0 +1,49 @@
+- Feature Name: Add point zero to major releases for Salt
+- Start Date: 2021-06-01
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Originally, Salt version numbers were date based as in YYYY.MM.DD and from SEP 14 we moved away from dates and general numbers, and removed the point zero from major release version numbers. This is to add that back to the major versions because it is problematic and inconsistent.
+
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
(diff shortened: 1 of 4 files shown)


max-arnold:
👍

salt-bootstrap needs to be fixed as well, to handle pinning to major versions. Some time ago I added a kludge for that: https://github.com/saltstack/salt-bootstrap/pull/1491

OrangeDog:
It's unclear. Are you proposing to go back and re-release e.g. 3000 as 3000.0?

If so, this could be even more disruptive. Un-releasing 3000 isn't really possible, and I'm sure there's at least one system that will treat 3000 and 3000.0 as different. More in-depth impact analysis is required.

If not, then it just makes things even more inconsistent. You can't put the genies of bad decisions back in the bottle once you've released them.

sagetherage:
I didn't intent to go back and update already released versions; however, if there is a reason to re-release I would be willing to discuss, but I did write it as to suggest as such. I can adjust. 

The intent of this proposal is add it back in as part of the process or standard operating procedure and to detail the decision going forward to include the point zero specifically because without it we have deviated from it and it causes many problems. I can include more details in the SEP.

bryceml (approved):

krionbsd (approved):

ScriptAutomate (approved):

s0undt3ch (approved):

garethgreenaway (approved):

cmcmarrow (approved):

waynew (approved):

PR #37, opened 2020-10-05
SEP 28: proposal to add inclusive-language

justindesilets:

PR #54, opened 2021-07-07
SEP 34: Inclusive Language Proposal

thatch45:
The Salt Project team has had many discussions around inclusive language. We would like to propose this SEP as a framework moving forward, and an explanation around preserving the term Salt-Master.

### DIFF
--- a/inclusive-language.md
+++ b/inclusive-language.md
@@ -82,6 +82,12 @@ to be started up via an alternative name. This alternative name will also be ava
 
 The "master" branch used in git will be changed to "main".
 
+## Interacting With Existing Systems
+
+Many systems currently exist which do not use inclusive language. In instances where the interaction or
+management of underlying systems does not use inclusive language, Salt will use the language of the underlying,
+managed system.
+
 ## Alternatives
 [alternatives]: #alternatives
 
--- a/inclusive-language.md
+++ b/inclusive-language.md
@@ -1,7 +1,7 @@
-- Feature Name: (inclusive-language)
-- Start Date: (2021-07-06)
+- Feature Name: SEP34: Inclusive Language
+- Start Date: 2021-07-06
 - SEP Status: Draft
-- SEP PR: (leave this empty)
+- SEP PR: https://github.com/saltstack/salt-enhancement-proposals/pull/54
 - Salt Issue: (leave this empty)
 
 # Summary
--- a/inclusive-language.md
+++ b/inclusive-language.md
@@ -1,0 +1,98 @@
+- Feature Name: (inclusive-language)
+- Start Date: (2021-07-06)
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The Salt Project has always sought to be inclusive and not offensive. Original designs and names used were done so
+with the intent of being inclusive. Over the last decade this landscape has evolved and certain changes need to be made.
+
+This SEP proposes how we can make a solid design that can be sustained as this landscape continues to evolve.
+
+This SEP is intended to supersede SEP #28
+
+# Motivation
+[motivation]: #motivation
+
+The use of inclusive language needs to be a continual discussion, but the project also needs to document
+why certain choices were made and the liabilities that come from specific changes to language.
+
+Understanding the intent derived from original language choices, as well as formalizing what have previously
+been informal policies will allow us to maintain a clean situation moving forward.
+
+# Design
+[design]: #detailed-design
+
+In 2020 a very valid surge of concern around inclusive language surfaced and has touched many aspects
+of society. It has also highlighted that our collective understating of inclusive language is
+evolving. The Salt Project has no intention of perpetuating language, practices, culture, or code
+that is offensive or non-inclusive.
+
+The evolving nature of the problem issues a concern on how to best manage this situation in the
+long term. The Salt Project is not staffed internally to adequately address these concerns. We
+also feel that the creation of a single static document to track this effort is not adequate.
+
+Fortunately, the primary sponsor of The Salt Project - VMware - maintains a significant effort to
+maintain inclusion and diversity. This SEP proposes that we strive to adhere to the guidelines
+presented by VMware for inclusive language.
+
+Details on the Diversity, Equity, and Inclusion efforts of VMware can be found here:
+
+https://www.vmware.com/company/diversity.html
+
+Fortunately, the Salt Project already adheres to the vast majority of rules set forth by
+the VMware Diversity, Equity, and Inclusion policies. Only a few things need to be addressed.
+
+Please note that these topics have been discussed at length and are actively being addressed.
+
+While we share a strong desire to adhere to inclusive language, we should not break user
+installs. We feel that if this initiative were to create added and onerous work, confusion, and/or
+frustration for our users then it could backfire in the overall intent to create a move understanding
+and peaceful world. This means that standard deprecation paths will be adhered to for changes and
+the impact of a change will need to be carefully considered.
+
+## Whitelist/Blacklist
+
+The primary change that needs to be made, and has already started to be executed, is to
+remove references to Whitelist and Blacklist.
+
+The usage of whitelist will be replaced with allowlist
+The usage of blacklist will be replaced with blocklist
+
+## Usage of the term "Master"
+Salt uses the term "Master" extensively to refer to the system which issues commands out to Salt Minions.
+I (Thomas Hatch) chose the term Master/Minion to refer to this process with the original intent of making
+the choice inclusive. The Salt Project team has extensively evaluated the usage of the term "Master" and has
+concluded that any attempt to change the term in its entirety would be deeply detrimental to our users.
+Due to the deep and extensive usage of the term "Master" we feel that an outright change would not be possible.
+Changing the term "Master" would break nearly every install of Salt worldwide and would therefore be detrimental
+to the project and, I feel, to the intent of the movement.
+
+We also feel that the term Master/Minion is non-derogatory due to the fact that the word Master is not paired
+with a derogatory term for a subservient entity. This was the original intent for the choice made in our verbiage.
+
+With this said, we feel that forcing the ongoing usage of the term "Master" in new documentation and presentation
+materials will potentially be an issue. Therefore we propose adding a new startup script allowing the Salt Master
+to be started up via an alternative name. This alternative name will also be used in introductory documentation
+and presentations as soon as possible.
+
+
+## Alternatives
+[alternatives]: #alternatives
+
+Many alternatives have been discussed previously and have been deemed untenable. These include maintaining
+a project level inclusiveness program as well as complete renames of the Salt Master. We feel strongly that
+these alternatives cannot be reliably executed given our available resources and goals.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+The main unresolved question of the alternative name that will be allowed for the Salt Master.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Drawbacks have been listed in this document for consideration.
--- a/inclusive-language.md
+++ b/inclusive-language.md
@@ -78,6 +78,9 @@ With this said, we feel that forcing the ongoing usage of the term "Master" in p
 materials will potentially be an issue. Therefore we propose adding a new startup script allowing the Salt Master
 to be started up via an alternative name. This alternative name will also be available for use in presentations.
 
+## Master Git Branch
+
+The "master" branch used in git will be changed to "main".
 
 ## Alternatives
 [alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


waynew (approved):

waynew:
I've expressed my frustration at being unable to find a replacement term for the Salt Master. Aside from that I'm :100: on board.

Terms we've considered and why I struggle to accept any of the other proposals:

- Leader is inadequate because the master doesn't lead - heck most of the time the Salt Master is going to have *less* responsibility than most of the minions.
- Director is better, but still falls short - it conveys someone that points the way, but doesn't specify how to get there. Also, in an organization one *can* ignore the director, sometimes even freely.
- Manager feels similarly ambiguous as a term.
- Controller is *okay*. 

A couple of interesting dictionary definitions that may also help describe my thoughts:


> From The Collaborative International Dictionary of English v.0.48 [gcide]:
> 
>   Master \Mas"ter\, v. t. [imp. & p. p. {Mastered}; p. pr. vb. n.
>      {Mastering}.]
>      1. To become the master of; to subject to one's will,
>         control, or authority; to conquer; to overpower; to
>         subdue.
>         [1913 Webster]
>   
>               Obstinacy and willful neglects must be mastered,
>               even though it cost blows.            --Locke.
>         [1913 Webster]
>   
>      2. To gain the command of, so as to understand or apply; to
>         become an adept in; as, to master a science.
>         [1913 Webster]

And the more wordy

> From Bouvier's Law Dictionary, Revised 6th Ed (1856) [bouvier]:
> 
>   MASTER. This word has several meanings. 1. **Master is one who has control** 
>   over a servant or apprentice. **A master stands in relation to his 
>   apprentices, in loco parentis, and is bound to fulfill that relation, which 
>   the law generally enforces. He is also entitled to be obeyed by his 
>   apprentices, as if they were his children**. Bouv. Inst. Index, h.t. 
>        2.-2. Master is one who is employed in teaching children, known 
>   generally as a schoolmaster; as to his powers, see Correction. 
>        3.-3. Master is the name of an officer: as, the ship Benjamin 
>   Franklin, whereof A B is master; the master of the rolls; master in 
>   chancery, &c. 
>        4.-4. By **master is also understood a principal who employs another to 
>   perform some act or do something for him**. The law having adopted the maxim 
>   of the civil law, qui facit per alium facit per se; **the agent is but an 
>   instrument**, and **the master is** civilly **responsible for the act of his agent**, 
>   as if it were his own, when he either commands him to do an act, or puts him 
>   in a condition, of which such act is a result, or by the absence of due care 
>   and control, either previously in the choice of his agent, or immediately in 
>   the act itself, negligently suffers him to do an injury.

(emphasis mine)

In my mind the Salt Master is one that has complete and utter control over the minions -- beyond just directing, leading, or managing. Controller, maybe even commander might be good enough...

I think of the Sorcerer's Apprentice, when Mickey creates some minions to do his bidding and they get away from him. When the master returns? He completely and utterly restores order.

Salt Master evokes this  mental picture (well, also Gru, because Master/Minions, but :joy: )

Salt Controller or Salt Commander not so much. That's my 2¢

thatch45:
I completely agree, it has been very difficult. I think that the main thing here is that we have a recognized alternative that can be used. I have heard many very strong arguments in favor of a change, but never a very good proposed name.

With that said, the goal of this SEP is not to chose a name, but to acknowledge the policy moving forward and document the difficulty in renaming Salt-Master. It is important that we have a document that covers the fact that a complete rename would be untenable.

bryceml (approved):

ScriptAutomate (approved):
I also prefer this superseding the more generic #37 proposal, as this SEP is specific to `salt` and the initiatives specifically driving/impacting the Salt Project.

The only suggestion I would have would be related to the branch naming scheme across repos coming out of Salt Project, going from `master` -> `main`, due to standards adopted by GitHub, GitLab, VMware, and others.

References:
- GitLab: https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/
- GitHub: https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/
  - And: https://github.com/github/renaming
- Renaming initiative, which enterprise companies such as VMware are involved in, and their section on `main`/`master`: https://inclusivenaming.org/language/implementation-path/#changing-the-git-branch-name

thatch45:
Good call @ScriptAutomate . I will get that in

OrangeDog:
What about other terms in VMWare policy? https://github.com/saltstack/salt-enhancement-proposals/pull/37#issuecomment-774251958
Will anything happen with these?
Will it depend on whether it's mirroring the language of the controlled system (e.g. `ps.kill_pid`, `system.shutdown_abort`, `apache_site.disabled`)?

ScriptAutomate:
> What about other terms in VMWare policy? [#37 (comment)](https://github.com/saltstack/salt-enhancement-proposals/pull/37#issuecomment-774251958)
> Will anything happen with these?
> Will it depend on whether it's mirroring the language of the controlled system (e.g. `ps.kill_pid`, `system.shutdown_abort`, `apache_site.disabled`)?

@OrangeDog Yeah, these are good points that have come up in conversation. Much of this has to do with Salt as the "glue" where we wouldn't want to ever replace the language used by the services, tools, applications, systems, etc. that Salt is being used to manage (otherwise, we would bring confusion and disconnect to users, and make our documentation more difficult to traverse).

This seems like it would be a good extension to:

> While we share a strong desire to adhere to inclusive language, we should not break user
installs. We feel that if this initiative were to create added and onerous work, confusion, and/or
frustration for our users then it could backfire in the overall intent to create a more understanding
and peaceful world. This means that standard deprecation paths will be adhered to for changes and
the impact of a change will need to be carefully considered.

Should this distinction, about Salt's purpose and Salt's need to adopt the language of the services it configures/manages/orchestrates/etc., be added to a paragraph following this one? I already like that the SEP already explicitly states only a few items, while openly stating that this would be an ongoing initiative that needs careful consideration. It would potentially make things clearer, for future ongoing efforts and explanations for why things like `ps.kill_pid` would continue to still exist, due to the underlying technologies that Salt is abstracting away. Especially for people driving initiatives in communities and orgs that are unfamiliar with Salt deployments, scripts/tools, underlying technologies, etc. in user environments.

dwoz (approved):

garethgreenaway (approved):

waynew (approved):

smokeytheblair (approved):

dmurphy18 (approved):

waynew:

---

To summarize the discussion:

A few changes have been made, this SEP addresses the inclusive language specific to Salt, as well as reconciling Salt to underlying systems. At this point the SEP is in Final Comment period and unless significant valid objections are made, given the existing approvals it should be merged after Wednesday, January 12, 2022.

tkwilliams:
May I point out the utter silliness of this entire effort?

I should say I grew up during the 60s and 70s in a very racially charged part of the US South, and even I find absolutely ZERO racial stigma associated with ANY of the terms on the VMWare list .  With effort, I suppose one could be offended by the term `master` -- IF IT WERE BEING APPLIED to a person, in implied authority over another person.  Anywhere after the Civil War (or I suppose guild apprenticeships if one is pedantic) one person being called the`master` of another is wrong, and offensive.  However, it is still a valid, well-defined, and meaningful English word, and using it for a process or network service which fulfills ALL the reasonable definitions of the term, and particularly in the absence of any better term (`overseer` anyone?  `commandant` perhaps?  those are jokes BTW if you can't tell) is just needless whitewashing.

The REST of the terms are flatly stupid (sorry, but there's no other word I can find) - there is NOTHING racially colored (or "non-inclusive" in the present terminology) about ANY of these terms:
- abort
- kill
- disable
- execute

`disable` is the MOST BIZARRE of them all to my eyes.  At least the others have some negative (that is, not "racial" or "non-inclusive", but at least NEGATIVE) connotations of SOME SORT (`abort` with regard to pro-choice/life arguments) and `kill`/`execute` of course being "violent") but what the HECK is bothersome to ANYONE about the word `disable`?  Asking seriously here, I really God's Honest Truth don't GET IT...

I should mention that I 1000% AGREE with renaming the `master` git branch to `main` - I've always considered `master` to be a stupid name for the main branch - it meets none of the criteria of a `master`.  It controls nothing, drives nothing.  Nothing reports to it, and it is responsible for no other branches.  I was fine when it was `develop` - that made SENSE, and although I didn't say anything when it switched to `master`, it was only because I missed the internal discussion beforehand, and once it was done it became a matter of "eh, whatever...".

But the current effort is just jaw-droppingly wasteful - SO MANY BETTER THINGS could be done with the cycles being wasted on this discussion.  When I think about the man-hours which will be burnt by the implementation of it, and THEN by all the end-users who have to retro-fit a zillion scripts, cronjobs, Jenkins jobs, RPM spec files, etc., etc., etc. ad nauseum, I honestly despair of people's priorities...

Sorry if anyone finds this offensive (seriously I really will be sorry if that happens) but I cannot sit quietly by without at least attempting to be the voice of common sense and reason in this thread...

Hugs all around,
Tom Williams

OrangeDog:
@tkwilliams FYI the initial git branch defaulted to “master” because it is the [master](https://www.thefreedictionary.com/master+copy) from which other copies are made. It has nothing to do with owning or controlling anything. The word has multiple definitions.

I assume “disable” is on the list because someone thinks it is offensive to people with disabilities

waynew:
@tkwilliams thanks for your input! These topics have been discussed at length, both in #37, and we've covered those points. Ultimately the goal is to have a more inclusive community, and where possible and reasonable, we'll make those changes.

Obviously there are terms like `kill` and `execute` that have a specific meanings within Linux and are unlikely to completely change.

For further reading, feel free to check out https://news.vmware.com/company/replacing-problematic-terminology

tkwilliams:
I'm sorry, but words are being used in a very unusual, non-standard, and JARGONISTIC fashion here.  "Inclusive" implies a value dichotomy of "inclusive" <-> "exclusive", where "inclusive" is granted some arbitrary positive value, while "exclusive" is negative.  That's fine - I can see that argument.

But the proposed VALUE MAPPING of random English words to be "inclusive" (good) vs. "exclusive" (bad) seem bizarre, random, and utterly arbitrary.  The fact that some random person may (or may not) be "bothered" by some random word, used entirely innocently and entirely within its correct definition is absolutely NUTSO.  The obvious reductio ad absurdum being that anyone anywhere and anytime can be "triggered" or "offended" by any random word they encounter.

This value mapping is particularly confusing when the words don't even IMPLICITLY "exclude" anyone AT ALL.  Exactly which group is excluded if I choose to "kill" a process, or "execute" a program?

The mother on "Dead Like Me" didn't like the word "moist" - does that imply that the word should be arbitrarily STRICKEN FROM THE LANGUAGE?  No, it implies that she had a hangup, that it was HER issue, not everyone else's, and she needed to get therapy, a clue, or whatever it was she needed to adjust to everyone else's perfectly reasonable language usage.

Other people's unreasonable "sensitivities" are NOT Holy Writ.  They're just that:  sensitivities, with no more validity or value than anyone ELSE'S desires to continue USING the words that accurately express what they wish to convey.  How is it that some people's likes and dislikes are SO MUCH MORE important than anyone else's?  This approach actually scans as extraordinarily EXCLUSIVE to me.   Wouldn't inclusiveness imply that one group's values don't violate or take precedence over another's?

As a specific case, "disable" (WRT services) implies "disable"/"enable", while "deactivate" pairs with "activate".  Need I point out that "activate" is less accurate and approriate term when used in this context - "activating" a service implies something much closer to "starting" than simply "enabling" and thus is much less appropriate for the functionality.

Lastly, I should mention that I am legally "disabled" (though I've never considered myself such at any point in my life), and I flatly did not see (and still honestly don't see, even after the above interpretation) that the VERB "disable" is in any way offensive, or even vaguely RELATED TO, the ADJECTIVE "disabled" as used to describe people.  I'm given to wonder whether ANY of the people who are PUSHING these changes are actually members of the supposed "offended minorities" or instead just well-meaning but misdirected folks performing feel-good whitewashing to make themselves feel better.  My personal experience is that these groups are 1) far less sensitive than imagined, and 2) much more concerned with results than fine grained semantic hair-splitting.

waynew:
There has been a lot of good discussion around inclusive language, both on this SEP and the one that it replaces. 

@tkwilliams you make an excellent point:

> [people are] much more concerned with results than fine grained semantic hair-splitting

We definitely are most interested in creating a welcoming, inclusive community. Just saying "oh we've added inclusive language so we're done" definitely *isn't* the end game - it's just part of making the Salt Project community more welcoming and inclusive.

And, as such, this SEP has been accepted, and we will work towards making these proposed changes as part of our on-going efforts to make and keep the Salt Project community an awesome place.

waynew:
Obviously this is too late to the party, but upon further knowledge after working on the syndics, I'm actually waaay less committed to master (the work/change/bug arguments still stand).

I was under the impression that the salt master was much more insistent about things but in the implementation details it's the minion that's insistent on reacting to messages from the salt master.

Anyway, neither here nor there, just wanted to throw that change of mind out for anyone who finds it useful.

PR #58, opened 2021-08-30
salt-ssh configurable compression proposal

nergdron:
This was recommended by @waynew on salt ticket [60479](https://github.com/saltstack/salt/issues/60479). It's a feature request to be able to configure the compression type and level used on thin tarballs for distribution by `salt-ssh`.

### DIFF
--- a/salt-ssh-configurable-compression.md
+++ b/salt-ssh-configurable-compression.md
@@ -1,0 +1,80 @@
+- Feature Name: salt-ssh-configurable-compression
+- Start Date: 2021-08-30
+- SEP Status: Draft
+- SEP PR:
+- Salt Issue: [#60479](https://github.com/saltstack/salt/issues/60479)
+
+# Summary
+[summary]: #summary
+
+we should be able to configure the type and strength of compression used on the
+salt thin tarball for distribution via salt-ssh.
+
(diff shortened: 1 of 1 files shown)


waynew:
This will be SEP 36 - feel free to update your filename/file contents :+1:

waynew:
:+1: from me. I think it's a great idea, and it seems like something that would be straightforward to implement. The main risk that I see is around the config, as you called out in the SEP. We definitely have native minions that would have different compression tools available, but if we could use some kind of targeting similar to the way we do with top files, in theory that should solve the problem.

It *may* be possible that we have to ship that information alongside the thin, but I'm not positive, having not looked closely at that code lately :upside_down_face:.

I think it would be 100% possible to build up this functionality in layers - if we start with just allowing support for any of the built-in compression, we could merge that in and then work up to whatever alternative approaches we need.

waynew (approved):

waynew:

---

To summarize:

No objections have been made, a few possible approaches were mentioned. Given the gap, after next week we'll go ahead and put this into final comment period.

dmurphy18 (approved):

whytewolf (approved):

PR #61, opened 2022-02-10
SEP 38: a stupid idea for a pillar fileserver.

whytewolf:
I know this isn't the most intelligent idea. and is not fully fleshed out currently. I welcome all criticisms. including the bad English.

### DIFF
--- a/0038-pillar_fileserver.md
+++ b/0038-pillar_fileserver.md
@@ -1,7 +1,7 @@
 - Feature Name: Pillar Fileserver 
 - Start Date: 2022-02-10
 - SEP Status: Draft
-- SEP PR:
+- SEP PR: 38
 - Salt Issue:
 
 # Summary
@@ -32,7 +32,7 @@ Here is a list of bugs/feature requests this would help remove.
 * [#61508](https://github.com/saltstack/salt/issues/61508)
 * [#59443](https://github.com/saltstack/salt/issues/59443) since the render engine is separated from the fetch engine would 
          all true env separation
-* 
+* [#14354](https://github.com/saltstack/salt/issues/14354)
 
 # Design
 [design]: #detailed-design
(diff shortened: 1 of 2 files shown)


waynew:
I probably need to read through this a few times but... I actually really like the fundamental idea happening here.

I've noticed as I've dug into several of the bits of Salt that there are places where we should just have a nice, well-defined API, and the backing service is just kinda like :shrug: cool whatever.

But we end out with some weird tweaks instead, for particular backing systems.

It would be pretty amazing, though, to have a more generic fileserver for pillars (and more?)

whytewolf:
Those points you made in https://github.com/saltstack/salt/issues/61508 are awesome for the git pfs, When i get some time I'll look into adding those into the draft, in part with examples of how to work with the fileserver from the backends perspective.

dmurphy18:
@waynew Thanks, thought - with Tiamat we can control the versions of pygit2/libgit2 used, so the issues encountered with the rpm/deb packages and the environments in use should be mitigated.  So the SEP might only apply for Tiamat based functionality which implies a parting of the ways from regular OS distributed packages.

whytewolf:
well this sep actually covers a lot more then just git_pillar. it separates out the file server fetch from the render completely. and adds an interface over top that lets any backend that works with files be used as a root pillar filesystem much like the normal salt fileserver. calling out the git functionality directly is a nice to have but it isn't the main focus. so much of the problems your describing with libgit and all that were never a consideration for this SEP.

This SEP was born more out of having 500 different ways of doing what amounts to the same thing with regards to handling pillar in different systems. forcing a splitting of tool functionality just to handle all the cases and not everything working together in a cohesive manner. 

that being said by the time it is implemented I am sure we will be on a tiamat build as the core system anyway. so I do expect the libgit issues to be covered by then. 

and as @waynew  pointed out this kind of starts down a solid path of removing hacks and starting to work more towards functional APIs. as well as allowing more entry points for salt extensions.

PR #62, opened 2022-02-17
Allow use YAML dictionaries instead of arrays in Salt SLS files

MurzNN:
This SEP is a continuation of tech debt issue https://github.com/saltstack/salt/issues/61649

### DIFF
--- a/XXXX-yaml-dictionaries.md
+++ b/XXXX-yaml-dictionaries.md
@@ -1,0 +1,168 @@
+- Feature Name: yaml-dictionaries
+- Start Date: 2022-02-17
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow use YAML dictionaries instead of arrays in Salt SLS files
+
+# Motivation
+[motivation]: #motivation
+
+Many people are embarassed when see that Salt uses arrays to describe key-value structures in SLS files, like this:
+```yaml
+/etc/foo.conf:
+  file.managed:
+    - source:
+      - 'salt://foo.conf?saltenv=dev'
+    - user: foo
+    - group: users
+    - mode: '0644'
+    - attrs: i
+```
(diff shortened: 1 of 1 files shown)


OrangeDog:
> @whytewolf: this actually would take a bunch of deep level changes. mainly to the way module.function is evaluated between high and low states.
-- https://github.com/saltstack/salt/issues/61649#issuecomment-1041873374

I don't think so. A renderer can just parse the yaml, then convert the dicts back to lists (e.g. via `.items()`).

The default `yaml` renderer could be changed to do it if it detects dicts, or a new `yaml_dict` renderer added.
People could do it right now with a custom renderer (e.g. in a salt-extension).

whytewolf:
> > @whytewolf: this actually would take a bunch of deep level changes. mainly to the way module.function is evaluated between high and low states.
> > -- [saltstack/salt#61649 (comment)](https://github.com/saltstack/salt/issues/61649#issuecomment-1041873374)
> 
> I don't think so. A renderer can just parse the yaml, then convert the dicts back to lists (e.g. via `.items()`).
> 
> The default `yaml` renderer could be changed to do it if it detects dicts, or a new `yaml_dict` renderer added. People could do it right now with a custom renderer (e.g. in a salt-extension).

the problem with the default yaml render getting involved at all,

the default yaml render is used a lot of places where that will change currently working behavior. such as pillar and default.yml. places were the returned information should be 1:1. in order to work with that the rendering would need to be context aware. which adds more processing of information into the renderer. 

one of the arguments was that this would improve performance but in having to make it context aware would actually decrease performance. 

to see an increase in performance you would have to do what i was saying and change the state behavior. 

however introducing a new renderer that is not contextual could be used to let users make the choice either way. default would still be the same and work in the same way. to enable the behavior described set `renderer: jinja|dict_yaml` in the minion. would still need to set your #! renders in default files.

OrangeDog:
(this isn't SEP 62 by the way, the SEP number is different to the PR number)

PR #64, opened 2022-03-30
Suggestion for a new way to write states

lowercase00:
Hi all.
Would like to propose a structural change in the way States are written.
Details on the .md file.
Thanks for opening this space up!


OrangeDog:
First impressions are I don't think it's any simpler. Maybe if you're already familiar with the complexity of k8s, but I'm not.

As with SEP-62, you can implement this yourself with a custom renderer. I'm not convinced it will handle every case as nicely as you imagine. Try `module.run` , extends, requisites, and other global state arguments.

lowercase00:
Thanks for your comment @OrangeDog ! 

Yes, I understand the possibility of implementing it with a custom renderer. But apart from the way we use and write states, this is a proposal mostly thinking of (bringing) new users in.

The explicit, "key first", approach (the dictionary key being descriptive) _tends_ to be more common across a range of tools (kubernetes, github actions, bitbucket pipelines, docker, ansible, stackstorm, drone.io and a few other tools we've used), and in my personal opinion, more readable, clearer and easier to write.

I understand that, as the current way of writing states has been around for quite a while, for existing (experienced and knowledgeable) Salt users this may come as unnecessary complexity. The goal for this specific proposal is to help making Salt simpler and more straight forward, and being a beginner myself I found it significantly harder to grasp Salt State than a few other of those tools mentioned.

The other aspect that I think could be relevant is internal adoption. For enterprise users that tend to have qualified and senior professionals in greater numbers, I imagine this tends to be less of an issue, but when thinking of adopting Salt internally on smaller companies, the perceived difficulty by junior team members as a whole end up being a relevant factor to consider, since training is (relatively speaking) more expensive and time consuming then on an enterprise with a fully qualified devops teams.

Obviously that's just how I interpreted things, and I understand that this may not be the case for the majority of users, these arguments are all hard (if not impossible) to measure / test / assess .

jcockhren:
I like the spirit of this SEP and can empathize with the author. I think the scope of this SEP should be reduced to focusing on metadata. 

I do not appreciate having to write extremely verbose YAML because it's harder for me to pick up and write/template (e.g., k8s). I like salt b/c the conventions allow me to write and test Salt states very quickly. However, the author has made a valuable point about having more descriptive keys would lead to more clarity for those who want to review/learn states.

I like the addition of a `metadata` key. This would need to be optional to ensure backward compatibility. I also think a state's `version` and `name` also belongs under the metadata key

```yaml
----

metadata:
  name: My awesome state
  description: This state does stuff
  version: v1
```

The metadata key alone would enhance states significantly when it comes to documentation. It would be trivial to auto-generate documentation using the `metadata` key.

On the other hand, I disagree the `spec` key format the author proposed:

- Forcing a user to manually create a unique ID for each state is burdensome.
- The entire file is a spec, so nesting under a `spec` key is unnecessary, in my opinion.
- an explicit `name` key is also unnecessary since the following already works:

```
install all the things:
  pkg.installed:
     - name: thethings 
```

I think the scope of this enhancement could be significantly reduced to get it across the finish line.

PR #71, opened 2023-06-22
SEP: Constrained Backends

monkz:
Constraining Backends to allow collaboration with unprivileged users.

### DIFF
--- a/0000-constrained-backends.md
+++ b/0000-constrained-backends.md
@@ -1,0 +1,110 @@
+- Feature Name: constrained backends
+- Start Date: 2023-06-23
+- SEP Status: Draft
+- SEP PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Apply constraints to individual backends to allow collaboration with "foreign" backends.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+This will open parts of the configuration for other developers (maybe in another department or even external) to modify, but without giving them root rights, even if the minion runs as root.
+Currently the publisher_acl is only applied via the salt api - but if a state enters the salt master via a (git) backend, there is no constraint checking, allowing anyone that can change an backend to basically change almost everything, if not everything.
+For example: This hinders to allow "foreign" git repositories to be used as backend without extensive merge approval process.
+
+In the end, an application administrator could contribute states / functions / ... that are specific to a software and enable more complex orchestrations, without having a merge approver to know everything about every application to spot (unintentional or intentional) malicious changes and an expert keeps being an expert for just one/assigned scope/s, without the need to be trusted with everything.
+
+# Design
+[design]: #detailed-design
+
+This is the bulk of the SEP. Explain the design in enough detail for somebody familiar
+with the product to understand, and for somebody familiar with the internals to implement. It should include:
+
+- Definition of any new terminology
+  - "foreign" backend: a backend that can be modified by other users than the salt master administrator.
+- Examples of how the feature is used:
+      * Target minions on backends
+        Users able to change webserver.git aren't always the same users that should be able to modify backupservers.
+      * Apply these states constrained to be applied as user x.
+        * This and more could be accomplished by forcing a render pipeline, that acts as gateway that applies or filters wrapping the renderers with another renderer, that applies/filters parts. Or simply raise an exception to stop further processing of a configuration state.
+      * Allow just a certain list of renderers.
+         * Some renderers  (like py) might have more capabilities than others to try to circumvent pillar scopes.
+
+
+      ```
+                gitfs_remotes:
+                  - https://foo.com/main.git
+                  - https://foo.com/webserver.git:
+                    - target: web-group-a\*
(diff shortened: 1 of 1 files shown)


0000-constrained-backends.md
```diff
@@ -0,0 +1,110 @@
...
+                    - renderer-allow:
+                      -  jinja|yaml
+                      -  py
+                  - https://foo.com/backups.git:
+                    - target: backup\*
+      ```
+  
+  The last part, is still unpolished:
+      * Render (some) state files in a sandbox.
+         * Rendering itself can be harmful on the salt master, as the renderer (for example: py) might get access to data / pillars that is not in scope for the targeted minion. Restricting the information that is available.
```
OrangeDog, line 61: This isn't true. State rendering happens on the minion. There's no way to access other minions' pillars.

0000-constrained-backends.md
```diff
@@ -0,0 +1,110 @@
...
+with the product to understand, and for somebody familiar with the internals to implement. It should include:
+
+- Definition of any new terminology
+  - "foreign" backend: a backend that can be modified by other users than the salt master administrator.
+- Examples of how the feature is used:
+      - Target minions on backends
+        Users able to change webserver.git aren't always the same users that should be able to modify backupservers.
+      - Apply these states constrained to be applied as user x.
+        - This and more could be accomplished by forcing a render pipeline, that acts as gateway that applies or filters wrapping the renderers with another renderer, that applies/filters parts. Or simply raise an exception to stop further processing of a configuration state.
+      - Allow just a certain list of renderers.
```
OrangeDog, line 36: How does this help? Every renderer can run arbitrary commands as `root`.

0000-constrained-backends.md
```diff
@@ -0,0 +1,110 @@
...
+
+This is the bulk of the SEP. Explain the design in enough detail for somebody familiar
+with the product to understand, and for somebody familiar with the internals to implement. It should include:
+
+- Definition of any new terminology
+  - "foreign" backend: a backend that can be modified by other users than the salt master administrator.
+- Examples of how the feature is used:
+      - Target minions on backends
+        Users able to change webserver.git aren't always the same users that should be able to modify backupservers.
+      - Apply these states constrained to be applied as user x.
```
OrangeDog, line 34: How? Fork a minion process running as the target user to run it? Most stuff would break, such as communication with the master.

0000-constrained-backends.md
```diff
@@ -0,0 +1,110 @@
...
+        - This and more could be accomplished by forcing a render pipeline, that acts as gateway that applies or filters wrapping the renderers with another renderer, that applies/filters parts. Or simply raise an exception to stop further processing of a configuration state.
```
OrangeDog, line 35: Changing the rendering has no effect on the running of the rendered states.

0000-constrained-backends.md
monkz: Sorry, my mixup. I corrected it.
Had originally in mind to call the minion a submaster of sorts.

OrangeDog: No, you've not corrected it. That's not how Salt works.

0000-constrained-backends.md
monkz: A wrapper can ensure that the input (the file from the file backend) and output of a rendering pipeline fits certain criteria. First scanning for arbitrary commands/functions/... in the state file and later checking for arbitrary commands/functions in the returned output.
Limiting the list of renderers should simply allow a salt administrator to concentrate on a limited set of features that should be allowed/denied/enforced - as you said every renderer can run arbitrary commands, but those can be issued with a particular format subset that a renderer can understand. I think, the py renderer would be excluded most of the time.

But you are correct for this, encapsulating a constrained renderer in a sandbox like environment would be favorable. Hence my suggestion to allow an salt administrator to sandbox the rendering within a renderer has only access to scoped data and limited acting room to have an effect on the host/minion.

0000-constrained-backends.md
monkz: "Target minions on backends" would modify (enable / disable) overlays of the fileserver on the base of targeting - which is controlled by the master. It would allow to create a "view" on a environment. This is maybe a feature on its own.

The latter "Apply these states constrained to be applied as user x" is tricky . as it would have to postprocess the output of a renderer, to ensure something like "runas" is set.
As for communication: The minion would have to act as proxy/filter for the unprivileged subprocess.

0000-constrained-backends.md
monkz: As the runtime can only run what it was given, i'll try to limit the output of the runtime by limiting the input of it.

0000-constrained-backends.md
monkz: https://github.com/monkz/salt-enhancement-proposals/commit/95946d9128cca982f0aff53128c38ef2cb5d7076

monkz: Seems like Github is not updating this PR with the latest commits i've pushed to my fork's master.

monkz: Found a solution for githubs lack of PR updates: https://github.com/orgs/community/discussions/16351#discussioncomment-6327770 Sorry for the notification noise.

